PR #58, opened 2022-03-11
rename android and try into android device, fix packages

ramirobg94:


android/app/BUCK
```diff
@@ -35,12 +35,12 @@ android_library(
 
 android_build_config(
     name = "build_config",
-    package = "com.arenamobile",
+    package = "org.openforis.collect.arena",
```
SteRiccio, line 38: should the package name be `arena` or `arena-mobile`?

ramirobg94: What do you prefer?

SteRiccio: I think Arena mobile should be OK, to avoid confusion with the desktop app.
But actually I don't know, if you have modified it already, keep it as it is, just arena, hoping that it won't be a problem in a near future with the other imported packages like arena-core, right?!

PR #62, opened 2022-03-17
read connection settings

ramirobg94:


src/state/app/reducers/ui.js
```diff
@@ -5,10 +5,19 @@ import initialState from '../initial.state';
...
       loading: isLoading,
     }),
+    [actions.setError]: (state, {payload: {error}}) => ({
+      ...state,
+      error: error,
+    }),
+    [actions.initConnection$]: state => ({
+      ...state,
+      error: false,
+      isLoading: true,
```
SteRiccio, line 19: is the variable in the state called loading or isLoading?!

ramirobg94: isLoading
I prefer to put is to booleans

SteRiccio: uhm, I prefer "loading" because isLoading lets met think to a function instead of a property... but I've asked this because at line 10 you used "loading" instead...!

PR #63, opened 2022-03-22
Fix cycle dependency screen

ramirobg94:
This fix a cycle dependency


src/screens/Home/screen.js
```diff
@@ -1,46 +1,25 @@
...
 import Header from 'arena-mobile-ui/components/Header';
 import Layout from 'arena-mobile-ui/components/Layout';
 import baseStyles from 'arena-mobile-ui/styles';
 import NavigateToSettings from 'navigation/components/NavigateToSettings';
-import {selectors as userSelectors} from 'state/user';
 
-import LoggedInAs from './components/LoggedInAs';
-import NotLoggedIn from './components/NotLoggedIn';
+import Survey from './components/Survey';
+import User from './components/User';
```
ramirobg94, line 10: It is going to be added into future commits, I will open a WIP pr to put all the surveys handling there

SteRiccio (approved):

src/screens/Home/screen.js
SteRiccio: I would rename the components into SurveyView or UserView just to avoid collisions with the Survey and User types in arena-core... But let's see when you will use them.

PR #65, opened 2022-03-25
Fix home

ramirobg94:


src/arena-mobile-ui/components/EmptyBlock/component.js
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import {View, Text} from 'react-native';
+
+import Button from 'arena-mobile-ui/components/Button';
+import Card from 'arena-mobile-ui/components/Card';
+import baseStyles from 'arena-mobile-ui/styles';
+
+import styles from './styles';
+
+const EmptyBlock = ({title, info, ctaLabel, onPress}) => {
```
SteRiccio, line 10: on Press what? On button press?

src/arena-mobile-ui/components/Button/index.js
```diff
@@ -5,13 +5,29 @@ import baseStyles from 'arena-mobile-ui/styles';
 
 import styles from './styles';
 
-const Button = ({onPress = null, label, type = 'primary', ...props}) => {
+const Button = ({
+  onPress = null,
```
SteRiccio, line 9: Can the default props go to the component defaultProps instead? And props here is actually otherProps or something like that...

SteRiccio (approved):

PR #67, opened 2022-04-05
Password input field: show/hide password with icon

SteRiccio:
To show the password show/hide icon inside the input field, I'm using a View that wraps the input field and the icon and giving to the View (almost) the same style as the "normal" input field.

### DIFF
--- a/src/arena-mobile-ui/components/PasswordInput/styles.js
+++ b/src/arena-mobile-ui/components/PasswordInput/styles.js
@@ -1,0 +1,22 @@
+import {StyleSheet} from 'react-native';
+
+import inputStyles from '../Input/styles';
+
+const styles = StyleSheet.create({
+  container: {
+    flex: 1,
+  },
+  inputContainer: {
+    ...inputStyles.input,
+    width: '100%',
+    flexDirection: 'row',
+    alignItems: 'center',
+    padding: 0,
+    paddingVertical: 2,
+  },
+  input: {
+    width: '95%',
+  },
+});
+
+export default styles;
--- a/src/arena-mobile-ui/components/PasswordInput/styles.js
+++ b/src/arena-mobile-ui/components/PasswordInput/styles.js
@@ -8,14 +8,14 @@ const styles = StyleSheet.create({
   },
   inputContainer: {
     ...inputStyles.input,
-    width: '100%',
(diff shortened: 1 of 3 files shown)


src/arena-mobile-ui/components/PasswordInput/styles.js
```diff
@@ -0,0 +1,22 @@
...
+    alignItems: 'center',
+    padding: 0,
+    paddingVertical: 2,
+  },
+  input: {
+    width: '95%',
+  },
+});
+
+export default styles;
```
ramirobg94, line 22: Maybe you can export as a component into the Input

export { default as Input, PasswordInput } ...

And compose the styles, ={[style.1, style.2]}
Also I prefer not to use %  in styles

src/screens/ConnectionSettings/screen.js
```diff
@@ -14,6 +14,7 @@ import Card from 'arena-mobile-ui/components/Card';
 import Header from 'arena-mobile-ui/components/Header';
 import Input from 'arena-mobile-ui/components/Input';
```
ramirobg94, line 15:
```suggestion
import Input, { PasswordInput } from 'arena-mobile-ui/components/Input';
```

src/arena-mobile-ui/components/PasswordInput/styles.js
SteRiccio: maybe it's better to keep them in separate files: Input could go to a Input.js file, PasswordInput to its own PasswordInput.js file and they could be both exported by an index.js file in Input. If I move it inside index.js, the style.js should be renamed and put in the same folder... a bit messy...

src/screens/ConnectionSettings/screen.js
SteRiccio: Why don't we export all the components from an index.js file in the components folder?

src/arena-mobile-ui/components/PasswordInput/styles.js
SteRiccio: composing styles is nice, but even the spread operator is allowed, and I think that the component should not be aware of how the style of the component is made up, so having it all in the styles.js file is better for me...

PR #71, opened 2022-04-22
Tests/node update italy

ramirobg94:
Pr with changes to start the nodeupdate,
important files are:

The code to implement is: 
src/state/nodes/sagas/updateNode/methods/index.js

The tests are here:
src/state/nodes/sagas/updateNode/__tests__/index.test.js

### DIFF
--- a/src/state/nodes/sagas/updateNode/__tests__/index.test.js
+++ b/src/state/nodes/sagas/updateNode/__tests__/index.test.js
@@ -1,0 +1,104 @@
+import MockDate from 'mockdate';
+import {expectSaga} from 'redux-saga-test-plan';
+
+import globalInitialState from 'state/initial.state';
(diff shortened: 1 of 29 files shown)


src/state/nodes/sagas/updateNode/methods/index.js
```diff
@@ -0,0 +1,92 @@
+const evalExpression = ({expression, node, record, nodes, nodeDefs}) => {
```
ramirobg94, line 1: this is the important file

src/state/nodes/sagas/updateNode/__tests__/index.test.js
```diff
@@ -0,0 +1,104 @@
+import MockDate from 'mockdate';
```
ramirobg94, line 1: this is the test file

PR #72, opened 2022-04-22
Test/survey

ramirobg94:


src/screens/Surveys/components/Remote/component.js
```diff
@@ -16,6 +16,12 @@ const Remote = ({
 }) => {
   const {loading, error, surveys = []} = surveysHooks.useRemoteSurveys();
 
+  useEffect(() => {
+    if (!loading && surveys.length <= 0) {
```
SteRiccio, line 20: Can the length be < 0? ===0 is enough I think

SteRiccio (approved):

src/screens/Surveys/components/Remote/component.js
ramirobg94: we can change if you want

PR #73, opened 2022-05-04
Feature/form

ramirobg94:


src/screens/ConnectionSettings/screen.js
```diff
@@ -69,14 +69,18 @@ const ConnectionSettings = () => {
...
+        const qrDataParsed = JSON.parse(qrData);
+
+        setFormData(prevData => ({
+          ...prevData,
+          username: qrDataParsed?.username || prevData.username,
+          serverUrl: qrDataParsed?.serverUrl || prevData.serverUrl,
+          password: qrDataParsed?.password || prevData.password,
+        }));
+      } catch (e) {
+        console.log('qrData is not a valid json');
```
SteRiccio, line 82: is there a better logging strategy we can use or put in place?

src/state/sagas.js
```diff
@@ -12,4 +13,5 @@ export default function* rootSaga() {
   yield all([fork(survey)]);
   yield all([fork(records)]);
   yield all([fork(nodes)]);
+  yield all([fork(form)]);
```
SteRiccio, line 16: is it necessary to use yield all even if the array param has only one element?

src/state/surveys/api/api.js
```diff
@@ -8,7 +8,9 @@ const getSurveys = async ({serverUrl}) => {
 
 const getSurveyById = async ({serverUrl, surveyId}) => {
   const {data: survey} = await API({serverUrl}).get({
-    path: `api/survey/${surveyId}`,
+    path: `api/survey/${surveyId}?${new URLSearchParams({
```
SteRiccio, line 11: maybe the parameters could be passed as an object to the API.get method, so the use of URLSearchParams could be moved there...

SteRiccio (approved):

src/screens/ConnectionSettings/screen.js
ramirobg94: Idk maybe console.error

src/state/sagas.js
ramirobg94: I used to initialize in this way maybe we can put togthr

src/state/surveys/api/api.js
ramirobg94: Yep I have to update the internal method to get params

src/screens/ConnectionSettings/screen.js
SteRiccio: maybe the user should be informed in this case, when something went wrong, with a toast or something like this

SteRiccio (approved):

PR #75, opened 2022-05-08
parser

ramirobg94:
Mock files for tests: 
src/arena/expression/_tests/mock/record.js (331 lines)
src/arena/expression/_tests/mock/survey.js (344 lines) 

TESTS:
src/arena/expression/_tests/parser.test.js (116)


src/arena/expression/parser.js
```diff
@@ -0,0 +1,117 @@
...
+      nodeDefsByUuid[node.nodeDefUuid].propsAdvanced?.validations?.expressions;
+  }
+
+  if (type === 'APPLICABLE') {
+    _expressions = nodeDefsByUuid[node.nodeDefUuid].propsAdvanced?.applicable;
+  }
+  if (type === 'DEFAULT_VALUES') {
+    _expressions =
+      nodeDefsByUuid[node.nodeDefUuid].propsAdvanced?.defaultValues;
+  }
```
ramirobg94: extract this to some strategy

src/arena/expression/parser.js
```diff
@@ -0,0 +1,117 @@
...
+        node,
+        survey,
+        record,
+      }),
+      jsApplyIf: _parseArenaExpression({
+        expression: expression.applyIf,
+        node,
+        survey,
+        record,
+      }),
```
ramirobg94: not sure about this "renaming", we should keep the expression and applyIf to create the error message if the error message is not defined

src/arena/expression/parser.js
```diff
@@ -0,0 +1,117 @@
...
+  });
+
+  return jsExpression; // Object.keys(nodeDefsUuidsByName);
+};
+
+export const prepareArenaExpressions = ({
+  node,
+  survey,
+  record,
+  type = 'VALIDATIONS',
```
ramirobg94: extract to types

src/arena/expression/parser.js
```diff
@@ -0,0 +1,117 @@
...
+    return `"${node.value}"`;
+  }
+  return node.value;
+};
+const _parseArenaExpression = ({expression, node, survey, record}) => {
+  const nodeDefsUuidsByName = getNodeDefUuidsByName({survey});
+  const nodeDefsByUuid = survey.nodeDefs;
+
+  let jsExpression = expression.replace();
+  /* HERE is where we should replace the MemberExpression and Identifiers with the 'raw' value  */
```
ramirobg94: we need to implement MemberExpressions, and Identifier Expressions to prepare the expression 
fromArena to JsExpression

src/arena/expression/parser.js
```diff
@@ -0,0 +1,117 @@
...
+      _node.parentUuid === null,
+  );
+};
+
+const getValueToExpression = ({node, survey}) => {
+  const nodeDefsByUuid = survey.nodeDefs;
+  const nodeDef = nodeDefsByUuid[node.nodeDefUuid];
+  if (nodeDef.type === 'text') {
+    return `"${node.value}"`;
+  }
```
ramirobg94: here we need to see what is the best way to transform Identifiers to Literals

src/arena/expression/parser.js
ramirobg94: functions like distance or categoryItemProp

src/arena/expression/parser.js
```diff
@@ -0,0 +1,117 @@
...
+    _node => _node.nodeDefUuid === nodeDefUuid,
+  );
+  const ancestorsNodeUuids = getNodeParents({node, record});
+
+  return nodesByNodeDefUuid.find(
+    _node =>
+      ancestorsNodeUuids.includes(_node.parentUuid) ||
+      _node.parentUuid === null,
+  );
+};
```
ramirobg94: this can be extracted to Survey and Record functions

src/arena/expression/parser.js
```diff
@@ -0,0 +1,117 @@
...
+      );
+    }
+  });
+
+  return jsExpression; // Object.keys(nodeDefsUuidsByName);
+};
+
+export const prepareArenaExpressions = ({
+  node,
+  survey,
```
ramirobg94: pass suevey it is not necessary, we need only the nodeDefs

src/arena/expression/parser.js
```diff
@@ -0,0 +1,117 @@
...
+  const nodeDefsByUuid = survey.nodeDefs;
+  const nodeDef = nodeDefsByUuid[node.nodeDefUuid];
+  if (nodeDef.type === 'text') {
+    return `"${node.value}"`;
+  }
+  return node.value;
+};
+const _parseArenaExpression = ({expression, node, survey, record}) => {
+  const nodeDefsUuidsByName = getNodeDefUuidsByName({survey});
+  const nodeDefsByUuid = survey.nodeDefs;
```
ramirobg94: maybe this can be part of the contexts of this parsers when we create the instance pass the survey and generate this "survey" context

src/arena/expression/_tests/mock/record.js
```diff
@@ -0,0 +1,331 @@
...
+    },
+    warnings: [],
+  },
+  nodes: {
+    CLUSTER_NAME_NODE_UUID: {
+      id: 33,
+      uuid: 'CLUSTER_NAME_NODE_UUID',
+      dateCreated: '2022-05-07T20:36:09.383Z',
+      dateModified: '2022-05-07T20:47:24.411Z',
+      parentUuid: 'CLUSTER_NODE_UUID',
```
SteRiccio: the cluster entity shouldn't have any parent

src/arena/expression/_tests/mock/record.js
```diff
@@ -0,0 +1,331 @@
+const RECORD = {
```
SteRiccio: wow... this is HUGE... why don't you use the RecordBuilder and SurveyBuilder exposed by arena core? a survey and a record built in that way is much easier to read...

src/arena/expression/_tests/mock/record.js
```diff
@@ -0,0 +1,331 @@
...
+    CLUSTER_NAME_NODE_UUID: {
+      id: 33,
+      uuid: 'CLUSTER_NAME_NODE_UUID',
+      dateCreated: '2022-05-07T20:36:09.383Z',
+      dateModified: '2022-05-07T20:47:24.411Z',
+      parentUuid: 'CLUSTER_NODE_UUID',
+      nodeDefUuid: 'CLUSTER_NAME_DEF_UUID',
+      value: 'bbbcdegf',
+      meta: {
+        h: ['CLUSTER_NODE_UUID'],
```
SteRiccio: strange that in the hierarchy there is an item... it's like the cluster has a parent entity...

src/arena/expression/_tests/mock/survey.js
```diff
@@ -0,0 +1,344 @@
+const SURVEY = {
```
SteRiccio: 300 lines of survey like this are simply impossible to understand

src/arena/expression/parser.js
```diff
@@ -0,0 +1,117 @@
+const getNodeDefUuidsByName = ({survey}) => {
```
SteRiccio: there is something similar in arena too, it's associated to the survey in a property called "nodeDefsIndex", managed by the surveyNodeDefsIndex module (index of the node definitions) but actually the index by node def name is missing there.
You should move this code to arena-core, in a module called SurveyWhateverYouWant. There is already a module called Surveys to read properties from a survey, I don't know if it's the right place.

src/arena/expression/parser.js
SteRiccio: some functions are quite similar to what in arena is managed by the recordNodesIndex module. You should move this code to arena-core and update arena and arena-mobile to use that.

src/arena/expression/parser.js
SteRiccio: of course, but this preparation of the context or the survey is something that should be done once, not every time we have to evaluate an expression. this is why in arena we used to have a "nodeDefsIndex" created (and associated to the survey) when the survey is fetched from the db.

src/arena/expression/parser.js
```diff
@@ -0,0 +1,117 @@
...
+  const nodeDefsByUuid = survey.nodeDefs;
+
+  let jsExpression = expression.replace();
+  /* HERE is where we should replace the MemberExpression and Identifiers with the 'raw' value  */
+  Object.keys(nodeDefsUuidsByName).forEach(nodeDefName => {
+    const hasNodeDef = jsExpression.includes(nodeDefName);
+    // not entity nodeDefs
+    if (
+      hasNodeDef &&
+      nodeDefsByUuid[nodeDefsUuidsByName[nodeDefName]].type !== 'entity'
```
SteRiccio: use the enum NodeDefType of arena-core

src/arena/expression/parser.js
SteRiccio: use SurveyDependencyType of arena-core

src/arena/utils.js
```diff
@@ -0,0 +1,27 @@
...
+  }
+  if (value.constructor === Set) {
+    return value.size === 0;
+  }
+  if (value.constructor === Map) {
+    return value.size === 0;
+  }
+
+  return false;
+};
```
SteRiccio: not another arena-core Objects module please...

SteRiccio (changes requested):

src/arena/expression/_tests/mock/record.js
ramirobg94: Ive just copy the state to do it faster, but agree

src/arena/expression/_tests/mock/record.js
ramirobg94: Cluster_name is an attribute of the cluster

src/arena/expression/_tests/mock/record.js
ramirobg94: Cluster name is an attribute of the cluster

src/arena/expression/_tests/mock/survey.js
ramirobg94: it is just a copy of the state to support the test, agree to build it with the builder

src/arena/expression/parser.js
ramirobg94: agree to moved that into XXX/Survey, I think that I put a comment above similar

src/arena/expression/parser.js
ramirobg94: Thats the idea with the comment, I did in this way because to "work" in "draft" is faster than iterate the module

src/arena/utils.js
ramirobg94: Ive just copy and paste util things to be faster than iterate with the arena-core module,

src/arena/expression/_tests/mock/record.js
SteRiccio: ah sorry, I tought it was the cluster entity

src/arena/utils.js
SteRiccio: you really like the arena-core module... :D

SteRiccio (approved):

SteRiccio (approved):

PR #76, opened 2022-05-08
Update node/expressions

ramirobg94:


src/arena/expression/parser.js
```diff
@@ -21,15 +21,27 @@ const getNodeParents = ({node, record}) => {
   return parentsUuids;
 };
 
+const getNodeSibilings = ({node, record}) => {
+  return Object.values(record.nodes)
+    .filter(
+      _node => _node.parentUuid === node.parentUuid && node.uuid !== _node.uuid,
+    )
+    .map(_node => _node.uuid);
+};
```
ramirobg94: this code is duplicated and it is also. into the selectors, the logic is going to be extracted to 
src/arena/core ( nor sure about the root )
to be a method of Records

src/state/nodes/sagas/updateNode/_tests/## Validations.pl
```diff
@@ -0,0 +1,67 @@
+## Validations
```
ramirobg94, line 1: this are the test to be implemented to check if the updateNode cover all the cases that we discussed, if you find some case that is not covered please add here

src/state/nodes/sagas/updateNode/methods/index.js
```diff
@@ -1,22 +1,23 @@
...
+const {
+  checkIfNodeExpressionsDependsOnReferenceNode,
+} = require('arena/expression/parser');
 
-const evalExpression = ({expression, node, record, nodes, nodeDefs}) => {
+const evalExpression = ({expression, node, record, survey}) => {
   // dode from arena -> const _getApplicableExpressions = (survey, record, nodeCtx, expressions, stopAtFirstFound = false) => {
 
-  return; //  error/warning(validationObject)
+  return node.value >= 0; //  error/warning(validationObject)
```
ramirobg94: here is going to be the evalExpr from jse-eval with the js expression ready

src/state/nodes/sagas/updateNode/methods/index.js
```diff
@@ -53,44 +67,71 @@ const updateDependantNodes = ({nodeDefs, node, record, nodes}) => {
...
+    record = {
+      ...record,
+      nodes: {
+        ...record.nodes,
+        ...(updatedNodes || []).reduce(
+          (acc, _node) => ({...acc, [_node.uuid]: {..._node}}),
+          {},
+        ),
+      },
+    };
```
ramirobg94: I feel dirty here but...

src/arena/expression/parser.js
SteRiccio: sure, and don't forget that iterating over all the nodes in a big record could be quiete slow, specially if you are comparing uuids... I don't want to say "in arena we did..." but in arena we have done it with an index (nodes by parent uuid) to simplify the search of children (at least)

src/arena/expression/parser.js
```diff
@@ -73,45 +85,116 @@ const _parseArenaExpression = ({expression, node, survey, record}) => {
...
-      ...expression,
-      jsExpression: _parseArenaExpression({
-        expression: expression.expression,
-        node,
-        survey,
-        record,
-      }),
-      jsApplyIf: _parseArenaExpression({
-        expression: expression.applyIf,
+  if (type === 'ALL') {
```
SteRiccio: what is this type "ALL"?

src/arena/expression/parser.js
```diff
@@ -73,45 +85,116 @@ const _parseArenaExpression = ({expression, node, survey, record}) => {
...
+    expression = [
+      ...(nodeDefsByUuid[node.nodeDefUuid].propsAdvanced?.validations
+        ?.expressions || []),
+      ...(nodeDefsByUuid[node.nodeDefUuid].propsAdvanced?.applicable || []),
+      ...(node.updatedBy !== 'USER'
```
SteRiccio: ?!

src/arena/expression/parser.js
```diff
@@ -73,45 +85,116 @@ const _parseArenaExpression = ({expression, node, survey, record}) => {
...
+  referenceNode,
+}) => {
+  const nodeDefsUuidsByName = getNodeDefUuidsByName({survey});
+  const nodeDefsByUuid = survey.nodeDefs;
+  const {expression, applyIf} = _expression;
+
+  let found = false;
+
+  Object.keys(nodeDefsUuidsByName).forEach(nodeDefName => {
+    const hasNodeDef =
```
SteRiccio: what is expression? a string? what if you have some other node def starting with "nodeDefName"?

src/arena/expression/parser.js
ramirobg94: Yep, we can also use [memoization](https://www.freecodecamp.org/news/understanding-memoize-in-javascript-51d07d19430e/) , better to delegate the responsibility of this Cache to the memory instead of us and have to store and recalculate again and again this "index"

src/arena/expression/parser.js
ramirobg94: What do you like to use to return all of the expressions?

src/arena/expression/parser.js
ramirobg94: how about to have updated_by into the nodes and now if the change was made by a user of the System ( default ) instead of having a flag that we have to maintain and clean, just a proposal open to discuss

src/arena/expression/parser.js
ramirobg94: in this case yes, 

This is how we have this info into the survey

```
{
    propsAdvanced: {
        validations: {
            expressions: [
                {
                    uuid: '1285f356-c0e2-4219-9c88-97c68f63fac7',
                    applyIf: '',
                    messages: {},
                    severity: 'error',
                    expression: 'cluster_name != "null"',
                },
            ],
        },
    defaultValues: [
        {
            uuid: '1536f442-1a1a-4bc7-8883-3392635d00e3',
            applyIf: 'cluster_key == 1',
            messages: {},
            severity: 'error',
            expression: '"aaa"',
        },
    ],

    applicable: [
        {
            uuid: 'e57d19cd-577a-4430-ae7b-a04e45c6194e',
            applyIf: '',
            messages: {},
            severity: 'error',
            expression: 'plot_key > 0',
        },
    ],
    }
}
```

What is the best way to name :

```
 {
                    uuid: '1285f356-c0e2-4219-9c88-97c68f63fac7',
                    applyIf: '',
                    messages: {},
                    severity: 'error',
                    expression: 'cluster_name != "null"',
},
```

and this?

```expression: 'cluster_name != "null"',```

File: src/arena/expression/parser.js (hunk 1 above)
Comment (SteRiccio) [reply src/arena/expression/parser.js]: I don't know what is faster actually... I think we have compared the 2 approaches (maybe 2 years ago...) with Mino: when you build the index you iterate all the nodes once, instead every call to a filter like the one you have implmeented iterates over all the nodes (and then is cached) but when the expression is evaluated it's done several times...

Comment (ramirobg94) [reply src/arena/expression/parser.js]: what approached did you try? store the index and meomization? or store the index and not store the index?

Comment (SteRiccio) [reply src/arena/expression/parser.js]: memoization without storing the index / generate and store the index: the second one was faster

File: src/arena/expression/parser.js (hunk 5 above)
Comment (SteRiccio) [reply src/arena/expression/parser.js]: pass a null type

File: src/state/form/selectors.js [hunk 8]
```diff
@@ -142,6 +142,7 @@ const getNodeDescendants = createCachedSelector(
   (recordNodes, node) => _getDescendants({nodes: recordNodes, node}),
 )((_state_, node) => node.uuid);
 
+// TO FIX nodeDef is not a condition to be a sibiling // also extraxt to Record and others functions using memoization
 const getNodeSibilings = createCachedSelector(
```
ramirobg94, line 146: getEntitySibilings, rename and add the correct one, also extract the internal logic to src/arena/core/Record/nodes

and use memoization

src/arena/expression/parser.js
SteRiccio: the default values are used even for the calculated attributes (read only attribute), they are not updated by the user

src/state/nodes/sagas/updateNode/index.js
```diff
@@ -27,16 +27,17 @@ function* handleUpdateNode({payload}) {
 
   //yield call(validateIfRootWithOtherRecords);
 
-  const {updatedNodes, validatedNodes} = yield call(updateNodeAndDependats, {
-    node,
-    updatedNode,
-    record,
-    recordNodes,
-    nodeDefsByUuid,
+  const recordWithNodes = {...record, nodes: {...recordNodes}};
```
ramirobg94, line 30: how about to use recordNodes instead of pass record all the time?

src/arena/expression/parser.js
SteRiccio: I was not talking about the name of the variable. One should be the expressionObject maybe, the other one expressionString... boh, anyway the problem is here: expression.includes(nodeDefName). If nodeDefName is "plot" and your expression is plot_id > 0, it will give you "true"

ramirobg94: I've put a comment above about this -> 
/* HERE is where we should replace the MemberExpression and Identifiers with the 'raw' value */

Bacause we should do with the Identifier and Member expressions parser

PR #88, opened 2022-05-30
add record list

ramirobg94:


src/screens/Records/components/ListRecords/component.js
```diff
@@ -3,17 +3,40 @@ import {View, Text} from 'react-native';
...
+    onSelect?.(record);
+  }, [record, onSelect]);
+
+  return (
+    <TouchableCard
+      onPress={handlePress}
+      customStyles={[styles.cardContainer, isSelected ? styles.selected : {}]}>
+      <View>
+        <Text>
+          {record.uuid} - {record.dateCreated}
```
SteRiccio, line 24: I think it's just a temporary solution to display the record, right?

SteRiccio (approved):

src/screens/Records/components/ListRecords/component.js
ramirobg94: yes

PR #95, opened 2022-06-05
Sagas/node

ramirobg94:


src/state/nodes/sagas/createNode/index.js
```diff
@@ -0,0 +1,77 @@
...
+
+  const childrenNodeDefs = yield select(state =>
+    surveySelectors.getNodeDefChildren(state, parentNodeDef),
+  );
+
+  if (childrenNodeDefs?.length <= 0) {
+    return [];
+  }
+
+  const nodes = yield all(
```
SteRiccio: What is this? Why don't you use arena core to create nodes? What about nodes in single entities and what about applying default values to that nodes?

src/state/nodes/sagas/createNode/index.js
```diff
@@ -0,0 +1,77 @@
...
+  const node = yield call(_createNode, {
+    record,
+    parentNode,
+    nodeDef,
+    survey,
+  });
+
+  yield put(nodesActions.setNode({node}));
+
+  if (nodeDef.type === 'entity') {
```
SteRiccio: Please use the NodeDefType in core

src/state/nodes/actionCreators.js
```diff
@@ -4,7 +4,10 @@ import types from './actionTypes';
 
 const {nodes} = createActions({
   [types.createNode$]: ({nodeDef}) => ({nodeDef}),
-  [types.updateNode$]: ({updatedNode}) => ({updatedNode}),
+  [types.updateNode$]: ({updatedNode, shouldClose}) => ({
+    updatedNode,
+    shouldClose,
```
SteRiccio: Should close what? Should close the node editor?

SteRiccio (changes requested):

src/state/nodes/sagas/createNodeAndDescendants/index.js
```diff
@@ -0,0 +1,41 @@
...
+  const recordWithNodes = {...record, nodes: {...(recordNodes || {})}};
+
+  const updateRecord = RecordNodesUpdater.addNodeAndDescendants({
+    survey,
+    record: recordWithNodes,
+    parentNode,
+    nodeDef,
+  });
+
+  // TODO Add surveyUuid into CORE and fix dates
```
ramirobg94: in core, because they are without timezone in core

SteRiccio (approved):

PR #99, opened 2022-06-06
delete node entiy in record

ramirobg94:


src/state/nodes/reducers/data.js
```diff
@@ -26,6 +26,15 @@ const data = handleActions(
       delete newState[node.uuid];
       return newState;
     },
+    [actions.deleteNodes]: (state, {payload: {nodes}}) => {
+      let newNodes = Object.assign({}, state);
```
SteRiccio: newNodes is never re-assigned, you could use const instead

SteRiccio (approved):

PR #114, opened 2022-06-22
Feature/showkeymessages

ramirobg94:


src/form/common/Validation/index.js
```diff
@@ -1,35 +1,76 @@
...
+        backgroundColor={configBySeverity.color}
         overlayColor={colors.transparent}
-        popover={<Text>Info here</Text>}>
-        <Icon name="warning-outline" color={colors.error} />
+        popover={
+          <>
+            {flatValidation.errors.map(error => (
+              <Text key={error.key}> {t(`Validation:${error.key}`)}</Text>
+            ))}
+            {flatValidation.warnings.map(warning => (
```
SteRiccio, line 68: Since they are rendered in the same way, maybe you can combine them into a single array and map that one

SteRiccio (approved):

src/form/common/Validation/index.js
ramirobg94: I am going to merge on the first and the fix both comments

PR #115, opened 2022-06-24
add decimal

ramirobg94:


src/form/Attributes/Decimal/Form/index.js
```diff
@@ -0,0 +1,13 @@
+import React from 'react';
+import {Platform} from 'react-native';
+
+import {Form} from '../../common/BaseInput';
+
+const IntegerForm = ({nodeDef}) => (
+  <Form
+    nodeDef={nodeDef}
+    keyboardType={(Platform.OS = 'ios' ? 'decimal-pad' : 'numeric')}
```
SteRiccio: is Patform.OS = 'ios' an assignment? Why don't you put these type of checks in a module like SystemUtils (or a better name) and you call a function like SystemUtils.isIOS()?

ramirobg94: good catch, Fixed in https://github.com/openforis/arena-mobile/pull/117, I will merge here all of them

SteRiccio (approved):

PR #119, opened 2022-06-27
Publish

ramirobg94:


.github/workflows/publish.yml
```diff
@@ -0,0 +1,54 @@
...
+        id: decode_keystore
+        uses: timheuer/base64-to-file@v1
+        with:
+          fileName: './android/arena-release-key.keystore'
+          encodedString: ${{ secrets.KEYSTORE }}
+      - name: Decode Play json
+        id: decode_play_json
+        uses: timheuer/base64-to-file@v1
+        with:
+          fileName: './android/4428354186-c469bd85b036.json'
```
SteRiccio: Is it possible to rename this file to a more meaningful name?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,55 @@
...
+        uses: timheuer/base64-to-file@v1.1
+        with:
+          fileName: 'arena-release-key.keystore'
+          fileDir: './android'
+          encodedString: ${{ secrets.KEYSTORE }}
+      - name: Decode Play json
+        id: decode_play_json
+        uses: timheuer/base64-to-file@v1.1
+        with:
+          fileName: '4428354186-c469bd85b036.json'
```
SteRiccio, line 45: can this file be renamed with a more meaningful filename? something like googl_play_key.json or whatever you want?

android/gradle.properties
```diff
@@ -26,3 +26,4 @@ android.enableJetifier=true
 
 # Version of flipper SDK to use with React Native
 FLIPPER_VERSION=0.99.0
+org.gradle.jvmargs=-Xmx4608M
```
SteRiccio, line 29: does it require all this memory? and why is it set to that strange value? should it be -Xmx4g? (4GB, that is a lot...)

src/state/records/selectors.js
```diff
@@ -18,8 +21,27 @@ const getRecordByUuid = createCachedSelector(
   (recordsByUuid, recordUuid) => recordsByUuid[recordUuid] || false,
 )((_state, recordUuid) => recordUuid);
 
+const gerRecordKey = createCachedSelector(
+  nodesSelectors.getNodesByRecordUuid,
+  surveyNodeDefsSelectors.getNodeDefRoot,
+  surveyNodeDefsSelectors.getNodeDefsByUuid,
+  (_, recordUuid) => recordUuid,
+  (nodes, nodeDefRoot, nodeDefsByUuid, recordUuid) => {
+    const rootNode = nodes.find(node => node.nodeDefUuid === nodeDefRoot.uuid);
```
SteRiccio, line 30: would it be better to check only that node.parentUuid is null? don't forget that there will be thousands of nodes per record (e.g. Naforma, ~4000 nodes per record) and comparing uuids is slow)

SteRiccio (approved):

PR #122, opened 2022-07-07
Design changes

ramirobg94:


src/form/Attributes/common/Base/Preview/index.js
```diff
@@ -1,12 +1,93 @@
...
+  return (
+    <View style={styles.container}>
+      <AttributeHeader nodeDef={nodeDef} nodes={nodes} />
+
+      {nodes?.map(node => (
+        <BasePreviewNode
+          key={node.uuid}
+          node={node}
+          nodeDef={nodeDef}
+          showValidation={nodes.length > 1}
```
SteRiccio, line 83: cant't you have one invalid node?

src/state/survey/selectors/categories.js
```diff
@@ -0,0 +1,106 @@
...
+  getRefData,
+  refData => refData?.categoryItemIndex,
+);
+
+const getNodeDefCategoryIndex = createCachedSelector(
+  getNodeDefByUuid,
+  getNodeDefsByUuid,
+  (nodeDef, nodeDefsByUuid) => {
+    let index = 0;
+    let parentCodeDef = nodeDef.props.parentCodeDefUuid;
```
SteRiccio, line 30: maybe it's better to rename parentCodeDef to parentCodeDefUuid, to avoid confusion

src/state/survey/selectors/categories.js
```diff
@@ -0,0 +1,106 @@
...
+  getRefData,
+  refData => refData?.categoryItemUuidIndex,
+);
+
+export const getCategoryItemIndex = createSelector(
+  getRefData,
+  refData => refData?.categoryItemIndex,
+);
+
+const getNodeDefCategoryIndex = createCachedSelector(
```
SteRiccio, line 25: this is not the category index, but the category level index

SteRiccio (approved):

src/form/Attributes/common/Base/Preview/index.js
ramirobg94: this validation appears into the input on the right it is only visible is there are multiple nodes, if not the validation appears into the header

PR #123, opened 2022-07-12
Feature/datetime

ramirobg94:


src/form/common/Form/component.js
```diff
@@ -109,4 +109,21 @@ const AttributeForm = () => {
...
+  if ([NodeDefType.date, NodeDefType.time].includes(nodeDef?.type)) {
+    return (
+      <>
+        {nodeDef &&
+          React.createElement(FormsByType[nodeDef?.type] || BaseForm, {
+            nodeDef,
+          })}
+      </>
+    );
+  }
```
SteRiccio, line 124: maybe you can do:
if (!nodeDef) return null
if ([NodeDefType.date, NodeDefType.time].includes(nodeDef?.type)) {
    return React.createElement...
}

SteRiccio (approved):

PR #125, opened 2022-07-12
boolean

ramirobg94:


src/form/Attributes/Boolean/Preview/index.js
```diff
@@ -0,0 +1,86 @@
...
+  );
+
+  return (
+    <View style={[styles.container]}>
+      <BooleanOption
+        label={t(
+          `Form:nodeDefBoolean.labelValue.${
+            nodeDef.props.labelValue || 'trueFalse'
+          }.true`,
+        )}
```
SteRiccio, line 65: the only difference between the 2 is the true/false value, can you wrap it into another inner component?

SteRiccio (approved):

PR #127, opened 2022-07-14
Feature/codeform

ramirobg94:


src/form/Attributes/Code/Preview/CodeCheckbox/index.js
```diff
@@ -0,0 +1,150 @@
...
+        event.preventDefault();
+
+        let _node = node;
+
+        if (NodeDefs.isMultiple(nodeDef) && Objects.isEmpty(_node)) {
+          _node = nodes.find(({value}) => Objects.isEmpty(value));
+        }
+        if (!NodeDefs.isMultiple(nodeDef) && Objects.isEmpty(_node)) {
+          _node = node || nodes[0];
+        }
```
SteRiccio, line 84: this code could be extracted to a function like "determineNode" or something like that

src/state/nodes/sagas/createNodeAndDescendantsWithValue/index.js
```diff
@@ -0,0 +1,28 @@
+import {call} from 'redux-saga/effects';
+
+import handleCreateNodeAndDescendants from '../createNodeAndDescendants';
+import handleUpdateNode from '../updateNode';
+
+function* handleCreateNodeAndDescendantsWithValue({payload} = {}) {
```
SteRiccio, line 6: createNodeAndDescendants makes sense when the node is an entity. This function should be called just something like "createAttributeWithValue"

src/state/nodes/sagas/removeNode/index.js
```diff
@@ -0,0 +1,36 @@
...
+  const [record, recordNodesByUuid] = yield all([
+    select(formSelectors.getRecord),
+    select(formSelectors.getRecordNodesByUuid),
+  ]);
+  const recordWithNodes = {...record, nodes: {...(recordNodesByUuid || {})}};
+
+  const recordUpdated = Records.removeNode(node)(recordWithNodes);
+
+  let nodesToDelete = [];
+  const newRecordsUuids = Object.keys(recordUpdated.nodes);
```
SteRiccio: newRecordsUuids is actually a newRecordNodesUuids

src/state/nodes/sagas/createNodeAndDescendantsWithValue/index.js
ramirobg94: at the end we are creating a node, and this function is used also with entities

src/state/nodes/sagas/removeNode/index.js
ramirobg94: true

src/state/nodes/sagas/createNodeAndDescendantsWithValue/index.js
```diff
@@ -0,0 +1,28 @@
...
+  const nodes = yield call(handleCreateNodeAndDescendants, {
+    payload: {
+      nodeDef,
+      parentNode,
+    },
+  });
+
+  const updatedNodes = Object.values(nodes);
+
+  const updatedNodeWithValue = Object.assign({}, updatedNodes[0], {value});
```
SteRiccio, line 18: assign the value only if specified (not necessary when creating an entity) or check if the node def is an attribute before doing this

src/state/nodes/sagas/createNodeAndDescendantsWithValue/index.js
SteRiccio: then call the function handleCreateNodeAndDescendants, or handleCreateNode... "with value" as I was saying applies only to attributes

src/state/nodes/sagas/removeNode/index.js
```diff
@@ -0,0 +1,36 @@
...
+  const recordUpdated = Records.removeNode(node)(recordWithNodes);
+
+  let nodesToDelete = [];
+  const recordUpdatedNodes = Object.keys(recordUpdated.nodes);
+  Object.keys(recordNodesByUuid).map(nodeUuid => {
+    if (!recordUpdatedNodes.includes(nodeUuid)) {
+      nodesToDelete.push(recordNodesByUuid[nodeUuid]);
+    }
+  });
+
```
ramirobg94, line 29:
```suggestion
// this code is needed if you dont have the index as arena-core search descendants by index
const currentDescentands = yield select(state =>
    formSelectors.getNodeDescendants(state, node),
  );

  nodesToDelete = nodesToDelete.concat(currentDescentands);
```

src/state/nodes/sagas/removeNode/index.js
```diff
@@ -0,0 +1,41 @@
...
+  const recordUpdated = Records.removeNode(node)(recordWithNodes);
+
+  let nodesToDelete = [];
+  const recordUpdatedNodes = Object.keys(recordUpdated.nodes);
+  Object.keys(recordNodesByUuid).map(nodeUuid => {
+    if (!recordUpdatedNodes.includes(nodeUuid)) {
+      nodesToDelete.push(recordNodesByUuid[nodeUuid]);
+    }
+  });
```
ramirobg94:
```suggestion
  Object.keys(recordNodesByUuid).forEach(nodeUuid => {
    if (!recordUpdatedNodes.includes(nodeUuid)) {
      nodesToDelete.push(recordNodesByUuid[nodeUuid]);
    }
  });
```

SteRiccio (approved):

PR #139, opened 2022-08-03
Taxonomies

ramirobg94:


src/form/Attributes/Taxonomy/Preview/index.js
```diff
@@ -0,0 +1,52 @@
+import React, {useCallback} from 'react';
+import {useSelector} from 'react-redux';
+
+import Select from 'arena-mobile-ui/components/Select';
+import useNodeFormActions from 'state/form/hooks/useNodeFormActions';
+import {selectors as surveySelectors} from 'state/survey';
+
+import {Preview as BasePreview} from '../../common/Base';
+
+const getCategoryItemLabel = ({item}) =>
```
SteRiccio: getCategoryItemLabel should be getTaxonItemLabel

src/form/Attributes/Taxonomy/Preview/index.js
```diff
@@ -0,0 +1,52 @@
...
+  `(${item.props.code}) ${item.props.genus}`;
+
+const Taxonomy = ({node, nodeDef}) => {
+  const actions = useNodeFormActions({nodeDef});
+
+  const items = useSelector(state =>
```
SteRiccio, line 16: all the taxonomy items (taxa) could be a long list (3000 items in some cases), maybe it would be better to use an autocomplete component in this case, to avoid rendering all the items in the select

src/form/common/Form/component.js
```diff
@@ -18,6 +18,7 @@ import DateForm from 'form/Attributes/Date/Form';
 import DecimalForm from 'form/Attributes/Decimal/Form';
 import FileForm from 'form/Attributes/File/Form';
 import IntegerForm from 'form/Attributes/Integer/Form';
+import TaxonomyForm from 'form/Attributes/Taxonomy/Form';
```
SteRiccio: maybe TaxonomyForm is not the best name. Taxonomy is the list of all the taxon, so TaxonForm could be enough

PR #141, opened 2022-08-08
Feature/import records

ramirobg94:


src/state/form/sagas/importRecords/index.js
```diff
@@ -0,0 +1,65 @@
...
+    delete recordData.nodes;
+  }
+  console.log('num nodes', Object.keys(nodes).length);
+
+  if (!Objects.isEmpty(recordData)) {
+    yield put(recordsActions.setRecord({record: recordData}));
+    const nodeObj = {};
+    Object.entries(nodes).forEach(([key, node]) => {
+      nodeObj[key] = Object.assign({}, node, {surveyUuid});
+    });
```
SteRiccio: since nodes is fetched remotely and you don't use it elsewhere, you could even do side effect on it, instead of recreating the nodes object.
You could do just:
 Object.values(nodes).forEach((node) => {
      node['surveyUuid'] = surveyUuid;
    });

src/state/records/api/api.js
```diff
@@ -1,3 +1,35 @@
-const recordsApi = {};
+import API from 'infra/api';
+
+const getRecords = async ({serverUrl, surveyId}) => {
+  try {
+    const res = await API({serverUrl}).get({
+      path: `api/survey/${surveyId}/records`,
+    });
+    if (res.status === 204) {
+      return true;
```
SteRiccio: should it return an empty array instead?!

src/state/records/api/api.js
```diff
@@ -1,3 +1,35 @@
...
+  }
+};
+
+const getRecord = async ({serverUrl, surveyId, recordUuid}) => {
+  try {
+    const res = await API({serverUrl}).get({
+      path: `api/survey/${surveyId}/record/${recordUuid}`,
+    });
+    if (res.status === 204) {
+      return true;
```
SteRiccio: why true? should it return null in this case?

SteRiccio (approved):

PR #171, opened 2022-09-07
Feat/save serverurl survey

### LINKED ISSUE #162: Handle multiserver Survey ( save url into survey )

ramirobg94:
fixes #162

### DIFF
--- a/src/state/surveys/sagas/index.js
+++ b/src/state/surveys/sagas/index.js
@@ -1,8 +1,8 @@
 import {StackActions} from '@react-navigation/core';
-import Toast from 'react-native-tiny-toast';
 import {takeLatest, put, select, call} from 'redux-saga/effects';
 
-import i18n from 'i18n';
+import {checkIfCurrentServerIsTheSurveysServer} from 'arena/survey';
+import {handleShowToast} from 'infra/toast';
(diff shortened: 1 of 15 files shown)


src/state/surveys/sagas/index.js
```diff
@@ -9,22 +11,48 @@ import {
   selectors as surveySelectors,
   actions as surveyActions,
 } from 'state/survey';
+import {selectors as surveysSelectors} from 'state/surveys';
```
SteRiccio, line 14: why don't you export "selectors" directly as "surveySelectors" ? It would be easier to import it with the auto-completion too...

src/state/surveys/sagas/index.js
```diff
@@ -9,22 +11,48 @@ import {
...
+          i18n.t('Surveys:toasts.server', {
+            surveyServer: currentSurvey?.serverUrl,
+            serverUrl,
+          }),
+          {
+            duration: 10000,
+          },
+        );
+        return;
+      }
```
SteRiccio: this could be moved to a separate function, like "checkServerUrl" or whatelse

SteRiccio:
for me it's ok, just a couple of comments

PR #177, opened 2022-09-16
Feat/update storage

ramirobg94:
TODO next PRs -> multicycle and use getRecordKey from arena-core


src/state/files/sagas/index.js
```diff
@@ -22,20 +21,29 @@ function* filterFileNodes(nodes) {
...
   const fileNodes = yield call(filterFileNodes, nodes);
 
   if (Object.keys(fileNodes).length > 0) {
-    const filesUuids = Object.values(fileNodes).map(
-      node => node?.value?.fileUuid,
-    );
-    yield call(handleDeleteFiles, filesUuids);
+    const filesByFileUuid = {};
+    Object.values(fileNodes).forEach(node => {
+      filesByFileUuid[node?.value?.fileUuid] = {
```
SteRiccio: here there is some risk that if the node or its value is null or undefined, you end up to put an undefined fileUuid in the filesByFileUuid object, isn't it? should you check it before adding it to the dictionary?

src/state/files/utils/index.js
```diff
@@ -1,12 +1,32 @@
 import * as fs from 'infra/fs';
 import {uuidv4} from 'infra/uuid';
+import {getSurveyFolder} from 'state/__persistence';
 
-const BASE_ARENA_FILES_PATH = `${fs.BASE_PATH}/arena-files`;
+const getFilesFolderPath = ({surveyUuid, cycle}) =>
+  `${getSurveyFolder(surveyUuid)}/${cycle}/files`;
+
+const getNodePath = ({surveyUuid, cycle, recordUuid, nodeUuid}) =>
```
SteRiccio, line 8: for every node you will have only one file, so basically you don't need this function but only getFilePath (typo there)

src/state/files/utils/index.js
```diff
@@ -1,12 +1,32 @@
...
     const fileUuid = node?.value?.fileUuid || uuidv4();
-    const destinationPath = `${BASE_ARENA_FILES_PATH}/${fileUuid}/${node?.value?.fileName}`;
+
+    const destinationPath = `${getfilePath({
+      surveyUuid: node.surveyUuid,
+      cycle: '0', // TODO when cycle
+      recordUuid: node.recordUuid,
+      nodeUuid: node.uuid,
+      fileUuid,
+    })}/${node?.value?.fileName}`;
```
SteRiccio, line 29: is it necessary to include the value.fileName in the path? isn't it enough to use just the fileUuid?

src/state/records/hooks/index.js
```diff
@@ -0,0 +1,90 @@
...
+  const keysWhitelist = [
+    'uuid',
+    'recordKey',
+    'dateCreated',
+    'dateModified',
+    'surveyUuid',
+    'cycle',
+  ];
+
+  let recordSummary = {};
```
SteRiccio: recordSummary could be a constant

src/state/records/hooks/index.js
```diff
@@ -0,0 +1,90 @@
...
+  const records = useSelector(surveySelectors.getRecords);
+
+  const getRecordsFromFs = useCallback(async () => {
+    if (surveyUuid && cycle) {
+      const recordsFiles = await getRecordsFiles({
+        surveyUuid,
+        cycle,
+      });
+
+      let _recordsByUuid = {};
```
SteRiccio: same here, you never re-assign recordsByUuid

src/state/records/sagas/deleteRecord/index.js
```diff
@@ -0,0 +1,36 @@
...
+
+function* handleDeleteRecord({payload}) {
+  const {recordUuid, callBack} = payload;
+  let record = yield select(state =>
+    recordsSelectors.getRecordByUuid(state, recordUuid),
+  );
+
+  const currentRecordUuid = yield select(formSelectors.getRecordUuid);
+  yield put(recordsActions.cleanRecord({recordUuid: recordUuid}));
+  if (record !== false && !Objects.isEmpty(record)) {
```
SteRiccio, line 20: why record !== false? is it a boolean or an object? it could be null or undefined... so checking if (record) should cover all the cases

src/state/records/sagas/deleteRecord/index.js
```diff
@@ -0,0 +1,36 @@
...
+  } else {
+    const cycle = yield select(surveySelectors.getSurveyCycle);
+    const surveyUuid = yield select(surveySelectors.getSelectedSurveyUuid);
+    record = {uuid: recordUuid, cycle, surveyUuid};
+  }
```
SteRiccio, line 25: I don't get it.. if the record does not exist, you create it and then you delete it?!

src/state/records/selectors.js
```diff
@@ -15,20 +15,24 @@ const _getAllRecordsList = createSelector(getRecordStateData, records =>
...
+);
+
 const getRecordByUuid = createCachedSelector(
   getRecordStateData,
   (_, recordUuid) => recordUuid,
   (recordsByUuid, recordUuid) => recordsByUuid[recordUuid] || false,
 )((_state, recordUuid) => recordUuid);
 
-const gerRecordKey = createCachedSelector(
+const getRecordKey = createCachedSelector(
```
SteRiccio, line 30: this kind of function belongs to the model (arena-core) and there are already some methods there to get the key attributes of an entity (or from the record using the root entity)

src/state/survey/sagas/uploadData/index.js
```diff
@@ -0,0 +1,213 @@
...
+const FILES_BASE_PATH = `${TMP_SURVEYS_BASE_PATH}/files`;
+
+function* handlePrepareRecordsData() {
+  try {
+    yield call(fs.mkdir, {dirPath: RECORDS_BASE_PATH});
+    const surveyUuid = yield select(surveySelectors.getSelectedSurveyUuid);
+    const cycle = yield select(surveySelectors.getSurveyCycle);
+
+    let recordFiles = yield call(getRecordsFiles, {surveyUuid, cycle});
+    recordFiles = recordFiles.filter(file => file.isFile());
```
SteRiccio: can getRecordsFile return directories instead of files?

SteRiccio (commented):
ok for me, I've added just some comments

src/state/files/utils/index.js
ramirobg94: could be, but you need the format. The file could be a doc, a wav, mp4, png, jpeg ... and this info is into the name

src/state/files/utils/index.js
ramirobg94: right, but when we export or import we use the fileuuid

src/state/records/selectors.js
ramirobg94: other PR

src/state/records/sagas/deleteRecord/index.js
ramirobg94: !Objects.isEmpty(record)) { check this

src/state/records/sagas/deleteRecord/index.js
ramirobg94: doesnt exist into the memory layer. But I need to delete from the Fs

src/state/files/utils/index.js
SteRiccio: isn't the fileName in the node value already?

src/state/records/sagas/deleteRecord/index.js
SteRiccio: yes, but I mean that you could write simply if (record && !Objects.isEmpty(record) ) {

src/arena/record/index.js
```diff
@@ -0,0 +1,10 @@
+export const getRecordKey = (nodes, nodeDefRoot, nodeDefsByUuid) => {
```
ramirobg94, line 1: next PRs se the arena-core function

src/state/files/utils/index.js
ramirobg94: we are taking from the node -> node?.value?.fileName

src/state/files/sagas/index.js
```diff
@@ -22,23 +21,40 @@ function* filterFileNodes(nodes) {
...
 
   if (Object.keys(fileNodes).length > 0) {
-    const filesUuids = Object.values(fileNodes).map(
-      node => node?.value?.fileUuid,
-    );
-    yield call(handleDeleteFiles, filesUuids);
+    const filesByFileUuid = {};
+    Object.values(fileNodes).forEach(node => {
+      if (node?.value?.fileUuid) {
+        filesByFileUuid[node?.value?.fileUuid] = {
```
SteRiccio, line 39: at this point you don't need to check it, you can do just node.value.fileUuid, or better, you can extract it to a const fileUuid and use that

src/infra/fs/index.js
```diff
@@ -60,21 +61,32 @@ export const mkdir = async (
...
+  }
+  return RNFS.mkdir(cleanPathWithBase(dirPath), options);
+};
 
 export const writeFile = async (
   {filePath, content, encoding} = {
     encoding: DEFAULT_ENCODING,
   },
-) => RNFS.writeFile(cleanPathWithBase(filePath), content, encoding);
+) =>
```
SteRiccio, line 76: you always call write file specifying the filePath and the content, I suppose, so the only optional parameter is "encoding"; this could be simply export const writeFile = async (
  {filePath, content, encoding = DEFAULT_ENCODING},
) 
then you don't need even to check later encoding || DEFAULT_ENCODING

src/state/survey/sagas/uploadData/index.js
```diff
@@ -0,0 +1,210 @@
...
+            uuid: _file.uuid,
+            props: {
+              ..._file.meta,
+              name: _file.meta.fileName,
+              nodeUuid: _file.nodeUuid,
+              recordUuid: _file.recordUuid,
+            },
+          },
+        ),
+      );
```
SteRiccio: why don't you do just:
```
filesArr.push(
          {
            uuid: _file.uuid,
            props: {
              ..._file.meta,
              name: _file.meta.fileName,
              nodeUuid: _file.nodeUuid,
              recordUuid: _file.recordUuid,
            },
         }
      );
```

SteRiccio (approved):

src/state/survey/sagas/uploadData/index.js
ramirobg94: true

PR #233, opened 2022-10-10
Enhancement/records list performance

### LINKED ISSUE #186: improve records screen performance

ramirobg94:
fixes #186

### DIFF
--- a/src/infra/fs/index.js
+++ b/src/infra/fs/index.js
@@ -76,6 +76,13 @@ export const writeFile = async ({filePath, content, encoding}) =>
     encoding || DEFAULT_ENCODING,
   );
 
+export const appendFile = async ({filePath, content, encoding}) =>
+  RNFS.appendFile(
+    cleanPathWithBase(filePath),
+    content,
+    encoding || DEFAULT_ENCODING,
+  );
+
 const getPathOfFile = filePath =>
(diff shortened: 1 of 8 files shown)


src/infra/fs/index.js
```diff
@@ -76,6 +76,13 @@ export const writeFile = async ({filePath, content, encoding}) =>
     encoding || DEFAULT_ENCODING,
   );
 
+export const appendFile = async ({filePath, content, encoding}) =>
+  RNFS.appendFile(
+    cleanPathWithBase(filePath),
+    content,
+    encoding || DEFAULT_ENCODING,
```
SteRiccio, line 83: DEFAULT_ENCODING could be specified as default in the function arguments

src/screens/Records/screen.js
```diff
@@ -16,7 +16,7 @@ const Records = () => {
   const [selectedRecord, setSelectedRecord] = useState(null);
 
   const resetSelectedRecord = useCallback(() => {
-    setSelectedRecord(null);
+    setSelectedRecord(false);
```
SteRiccio, line 19: I would put "null" instead of false... because the type of selected record is not boolean but object

SteRiccio (approved):

PR #234, opened 2022-10-11
hide templates

### LINKED ISSUE #222: -we should be not able to select published Templates as surveys -> remove templates from the lists

ramirobg94:
fixes #222

### DIFF
--- a/src/state/surveys/hooks/index.js
+++ b/src/state/surveys/hooks/index.js
@@ -17,9 +17,12 @@ export const useRemoteSurveys = () => {
       setError(false);
       setLoading(true);
       const _surveys = await apiSurveys.getSurveys({serverUrl});
+
       setSurveys(
         _surveys
-          .filter(survey => survey.status === 'PUBLISHED')
+          .filter(
+            survey => survey.status === 'PUBLISHED' && survey.template !== true,
+          )
           .sort(
             (sa, sb) => -moment(sa.dateModified).diff(moment(sb.dateModified)),
           ),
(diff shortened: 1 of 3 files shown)


src/state/surveys/hooks/index.js
```diff
@@ -17,9 +17,12 @@ export const useRemoteSurveys = () => {
       setError(false);
       setLoading(true);
       const _surveys = await apiSurveys.getSurveys({serverUrl});
+
       setSurveys(
         _surveys
-          .filter(survey => survey.status === 'PUBLISHED')
+          .filter(
+            survey => survey.status === 'PUBLISHED' && survey.template !== true,
```
SteRiccio, line 24: if you fetched the surveys with template=false, do you have to filter them even here?

ramirobg94: seems that not, but this issue was raised during the Team testing and the backend has the following:
Captura de Pantalla 2022-10-13 a las 14 59 31

So, not sure if the server is filtering well, I prefer to do double check here

SteRiccio: that endpoint is used by Arena too, to show the list of surveys or templates, so assume that it works because it has to.

SteRiccio: even status filtering is not needed: surveys never published won't be included.

ramirobg94: So I don't know why this issue was raised the other day during the team testing, maybe we can close this PR

SteRiccio: I think there were some surveys cloned from a template, the label of that surveys was misleading... templates are already excluded in the list of surveys! so yes, you can close it.

PR #236, opened 2022-10-13
fix password clean

### LINKED ISSUE #237: I am not able to clean the password

ramirobg94:
Fixes #237

### DIFF
--- a/src/screens/ConnectionSettings/screen.js
+++ b/src/screens/ConnectionSettings/screen.js
@@ -1,3 +1,4 @@
+import {Objects} from '@openforis/arena-core';
 import React, {useCallback, useEffect, useState} from 'react';
 import {useTranslation} from 'react-i18next';
 import {
@@ -36,9 +37,10 @@ const ConnectionSettings = () => {
   const [formData, setFormData] = useState({});
 
   const onChangeText = useCallback(
-    key => value => {
-      setFormData(prevValue => ({...prevValue, [key]: value}));
-    },
+    key =>
+      (value = '') => {
+        setFormData(prevValue => ({...prevValue, [key]: value}));
+      },
     [],
   );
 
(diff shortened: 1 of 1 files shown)


src/screens/ConnectionSettings/screen.js
```diff
@@ -177,13 +179,23 @@ const ConnectionSettings = () => {
...
+                  }
                   autoCapitalize="none"
                 />
                 <PasswordInput
                   title={t('ConnectionSettings:access_info_fields.password')}
                   onChangeText={onChangeText('password')}
-                  value={formData?.password || password}
+                  value={
+                    Objects.isEmpty(formData?.password) &&
+                    formData?.password !== ''
```
SteRiccio: this condition should be always false... am I wrong?!

src/screens/ConnectionSettings/screen.js
```diff
@@ -177,13 +179,23 @@ const ConnectionSettings = () => {
                 <Input
                   title={t('ConnectionSettings:access_info_fields.username')}
                   onChangeText={onChangeText('username')}
-                  value={formData?.username || username}
+                  value={
+                    Objects.isEmpty(formData?.username) &&
+                    formData?.password !== ''
```
SteRiccio: since you have used Object.isEmpty above, why don't you use it even at line 184?

ramirobg94:
```suggestion
                    formData?.username !== ''
```

src/screens/ConnectionSettings/screen.js
ramirobg94: a | Objects.isEmpty | !== '' |  && 
-- | -- | -- | -- |
undefined or null | True | True | True 
'' | False | False | False 
'xxxx' | False | True | False

The first row happens when you has an empty value at the beginning

SteRiccio: Objects.isEmpty('') is true

ramirobg94: okey i will change

SteRiccio (approved):

PR #246, opened 2022-10-21
Fix/code label

### LINKED ISSUE #200: Fix code label selector

### LINKED ISSUE #201: Keys when code appears like [Object,Object] and not as string on selectors

ramirobg94:
fixes #200 
fixes #201

### DIFF
--- a/src/arena/record/index.js
+++ b/src/arena/record/index.js
@@ -1,7 +1,10 @@
+const DEFAULT_JOIN_STRING = ',';
+const DEFAULT_STRING = '';
+
 export const getKeyNodeAsString = ({
   node,
   categoryItemIndex,
-  defaultString = '',
+  defaultString = DEFAULT_STRING,
 }) => {
   if (node?.value?.itemUuid) {
     const code = categoryItemIndex?.[node?.value?.itemUuid]?.props?.code;
@@ -13,8 +16,8 @@ export const getKeyNodeAsString = ({
 export const getKeyNodesAsString = ({
   nodes = [],
   categoryItemIndex,
-  joinString = ',',
-  defaultString = '',
+  joinString = DEFAULT_JOIN_STRING,
+  defaultString = DEFAULT_STRING,
 }) =>
   nodes
     .map(nodeKey =>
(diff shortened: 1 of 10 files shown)


src/arena/record/index.js
```diff
@@ -1,12 +1,74 @@
...
+  nodes,
+  nodeDefsByUuid,
+  categoryItemIndex,
+}) => {
+  const keyNodes = nodes.filter(
     node =>
-      node.parentUuid === rootNode.uuid &&
+      node.parentUuid === entity.uuid &&
       nodeDefsByUuid[node.nodeDefUuid].props.key,
   );
```
SteRiccio: you could call getKeyNodesForEntity instead of repeating it

src/arena/record/index.js
```diff
@@ -1,12 +1,74 @@
...
+    nodes,
+    nodeDefsByUuid,
+  });
+
+  return getKeyNodesAsString({
+    nodes: keyRootNodes,
+    defaultString: '-',
+    joinString: '/',
+    categoryItemIndex,
+  });
```
SteRiccio: what if you do just:
`getKeyNodesForEntityAsString({ entity: rootNode, ...})` and you pass some other parameter for the separator ecc?

src/form/Attributes/Code/Preview/hooks/useCode.js
```diff
@@ -1,16 +1,18 @@
...
-    ? `(${categoryItem?.props?.code}) ${categoryItem?.props?.labels?.[language]}`
-    : '-';
+const getCategoryItemLabel =
+  nodeDef =>
+  ({categoryItem, language}) =>
+    categoryItem?.props?.code
+      ? `(${categoryItem?.props?.code}) ${
+          categoryItem?.props?.labels?.[language] || ''
+        }`
+      : '-';
```
SteRiccio, line 13: you should consider the Show code prop in the nodeDef too

src/form/Attributes/Code/Preview/hooks/useCode.js
```diff
@@ -1,16 +1,18 @@
...
 import surveySelectors from 'state/survey/selectors';
 
-const getCategoryItemLabel = ({categoryItem, language}) =>
-  categoryItem?.props?.code
-    ? `(${categoryItem?.props?.code}) ${categoryItem?.props?.labels?.[language]}`
-    : '-';
+const getCategoryItemLabel =
+  nodeDef =>
+  ({categoryItem, language}) =>
+    categoryItem?.props?.code
```
SteRiccio, line 9: the code is mandatory... only categoryItem could be null

SteRiccio (approved):

src/form/Attributes/Code/Preview/hooks/useCode.js
ramirobg94: Thanks! 
There is one PR for that, I will add them

src/arena/record/index.js
ramirobg94: good catch I did the refactor and I didnt notice that

PR #267, opened 2022-11-10
Feat/code searchable

ramirobg94:


src/form/Attributes/Code/Form/index.js
```diff
@@ -1,5 +1,162 @@
...
+import formSelectors from 'state/form/selectors';
+
+import {useCode} from '../Preview/hooks';
+
+import styles from './styles';
+
+const ChevronDown = <Icon name="chevron-down" />;
+
+const getTextForSearch = item => {
+  const labels = Object.entries(item?.props?.labels || {}).map(
```
SteRiccio: I'm not sure if it's a good idea to search over all the labels... Also, in this case you could use Object.values instead of Object.entries

SteRiccio (approved):

SteRiccio (approved):

PR #274, opened 2022-11-15
Feat/multiple code searchable

ramirobg94:


src/form/Attributes/Code/Form/common/hooks/useSearch.js
```diff
@@ -0,0 +1,31 @@
+import {useState, useCallback} from 'react';
+
+export const getTextForSearch = item => {
+  const labels = Object.entries(item?.props?.labels || {}).map(
```
SteRiccio, line 4: I don't think it's a good idea to search over all the labels in all the languages: if you type something that matches a label in a language different from the current one, you will see the matching item in the list but the label shown will be another (the one in the current language)... Isn't it?!

SteRiccio (approved):

src/form/Attributes/Code/Form/common/hooks/useSearch.js
ramirobg94: changed into the https://github.com/openforis/arena-mobile/pull/267

PR #290, opened 2022-11-18
default value

### LINKED ISSUE #270: submit without change form saves null
cambodia_arena_test
> cluster inventory year

when a wrong entry (2014), and when in edit mode click “Save and return again”, value will disappear

ramirobg94:
fixes #270

### DIFF
--- a/src/form/Attributes/common/BaseInput/Form/index.js
+++ b/src/form/Attributes/common/BaseInput/Form/index.js
@@ -30,12 +30,14 @@ const Form = ({nodeDef, keyboardType = 'default'}) => {
 
   const handleSubmit = useCallback(
     ({callback = () => null} = {}) => {
+      const _newValue = newValue || String(node.value || '');
+
       const _value =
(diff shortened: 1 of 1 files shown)


src/form/Attributes/common/BaseInput/Form/index.js
```diff
@@ -34,8 +34,8 @@ const Form = ({nodeDef, keyboardType = 'default'}) => {
         nodeDef.type === NodeDefType.text
           ? transform({
               transformFunction: NodeDefs.getTextTransform(nodeDef),
-            })(newValue)
-          : Number((newValue || '').replace(',', '.'));
+            })(newValue || String(node.value || ''))
+          : Number((newValue || String(node.value || '')).replace(',', '.'));
```
SteRiccio: I see some duplication here... Can it be extracted to a variable to make this part more clear?

SteRiccio (approved):

PR #298, opened 2022-11-24
dont show language selector when only one

ramirobg94:


src/screens/Survey/components/SurveyLanguageSelector/component.js
```diff
@@ -27,6 +28,14 @@ const SurveyLanguageSelector = () => {
     [surveyLanguage, dispatch],
   );
 
+  if (surveyLanguages.length <= 1) {
```
SteRiccio, line 31: <= 1?! can it be 0??

SteRiccio (approved):

src/screens/Survey/components/SurveyLanguageSelector/component.js
ramirobg94: it happened with some surveys, I don't know if it was some punctual thing with some collect imported survey, in any case... I think that it is not bad to be defensive.

PR #310, opened 2022-11-30
sort surveys into the list

### LINKED ISSUE #221: -List of surveys: needed methods to sort them -> by: name, or date ( already there but keep it, use name but not label as the label could be on different languages)

ramirobg94:
resolves #221

### DIFF
--- a/src/screens/Surveys/components/common/List/component.js
+++ b/src/screens/Surveys/components/common/List/component.js
@@ -1,9 +1,12 @@
-import React, {useCallback} from 'react';
+import {Objects} from '@openforis/arena-core';
+import moment from 'moment';
+import React, {useCallback, useMemo} from 'react';
 import {useSelector} from 'react-redux';
 
 import CommonList from 'arena-mobile-ui/components/List';
 import {selectors as surveySelectors} from 'state/survey';
 
+import {SORTERS_DIRECTION, SORTERS_KEYS} from '../../Sorter/config';
 import SurveyCard from '../SurveyCard';
 const List = ({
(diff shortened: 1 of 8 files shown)


src/screens/Surveys/components/common/List/component.js
```diff
@@ -28,9 +32,31 @@ const List = ({
     [showIcons, selectedSurvey, setSelectedSurvey, localSurvey],
   );
 
+  const dataSorted = useMemo(() => {
+    if (Objects.isEmpty(sortCriteria)) {
+      return data;
+    }
+
+    if ([SORTERS_KEYS.dateModified].includes(sortCriteria?.type)) {
```
SteRiccio: What changes here is only the sort function, so you can define a sortFunctionByType or whatever and pass the function to data.sort

src/screens/Surveys/components/common/List/component.js
```diff
@@ -28,9 +32,31 @@ const List = ({
...
+      return data.sort(
+        (sa, sb) =>
+          moment(sa.dateModified).diff(moment(sb.dateModified)) *
+          (sortCriteria?.direction === SORTERS_DIRECTION.asc ? -1 : 1),
+      );
+    }
+    if ([SORTERS_KEYS.name].includes(sortCriteria?.type)) {
```
SteRiccio: You checked already above that sortCriteria is not empty... No need to check it again

SteRiccio (approved):

SteRiccio (approved):

PR #384, opened 2023-02-25
Surveylist

ramirobg94:


src/screens/Surveys/components/SurveysList/SubPanel/SurveysOriginSelector/component.js
```diff
@@ -0,0 +1,36 @@
...
+      return {
+        label: t('Surveys:subpanel.surveys_origin_selector.local.label'),
+        onPress: () => setSurveysOrigin('remote'),
+      };
+    } else {
+      return {
+        label: t('Surveys:subpanel.surveys_origin_selector.remote.label'),
+        onPress: () => setSurveysOrigin('local'),
+      };
+    }
```
SteRiccio: the code looks very similar, you could use the surveyOrigin to create the label key; same for the "new" origin: it is the other one
something like:
`const labelKey = ``Surveys:subpanel.surveys_origin_selector.${surveyOrigin}.label``
const newOrigin = surveyOrigin === 'local'? 'remote': 'local'`

SteRiccio (approved):

PR #391, opened 2023-03-13
logs

ramirobg94:


src/infra/objectUtils.js
```diff
@@ -14,9 +15,11 @@ export const deleteValueByKey =
 export const mergeSpread = (obj, newObj) => ({...obj, ...newObj});
 export const mergeShallow = (obj, newObj) => Object.assign({}, obj, newObj);
 export const mergeNoSpread = (obj, newObj) => {
+  perfState.start('mergeNoSpread');
   const result = Object.assign({}, obj);
   Object.keys(newObj).forEach(key => {
```
SteRiccio, line 20: here it would be better to iterate over the "entries", so you can save accessint to newObj[key]...

PR #392, opened 2023-03-17
New libraries

ramirobg94:


src/state/__persistence/index.js
```diff
@@ -65,7 +66,7 @@ export function* persistRecordWithKeyAndMergeCurrentNodes({record}) {
     );
     recordKey = yield call(
       getRecordKey,
-      Object.values(Objects.isEmpty(nodes) ? record.nodes : nodes),
+      Object.values(Objects.isEmpty(nodes) ? record.nodes || [] : nodes),
```
SteRiccio: record.nodes is an object of nodes indexed by uuid, not an array: you should default it with {}

SteRiccio (approved):
Good for me

src/state/__persistence/index.js
ramirobg94:
```suggestion
      Object.values(Objects.isEmpty(nodes) ? record.nodes || {} : nodes),
```

PR #415, opened 2023-04-20
Taxonomy settings

ramirobg94:


src/arena/taxonomy/index.js
```diff
@@ -0,0 +1,125 @@
+const props = {
+  uuid: 'uuid',
+  code: 'code',
+  scientificName: 'scientificName',
+  vernacularNames: 'vernacularNames',
+  genus: 'genus',
+  family: 'family',
+};
+
+export const taxonomyVisibleFieldsOptions = [
```
SteRiccio: here I see a lot of repetition.
The possible options can be an object like: 
{
    codeScientificName: [props.code, props.scientificName],
   scientificNameVernacularNames: [props.scientificName, props.vernacularNames],
   ...
}

then the "options" can be generated in the UI component (src/screens/Settings/Screens/Survey/Taxonomies/screen.js) by simply iterating over the entries of the taxonomyVisibleFieldsOptions or whatever you want to call it:
Object.entries(taxonomyVisibleFieldsOptions).map(([key, visibleFields]) => ({
   key,
   value: key,
  label: visibleFields.map(field => t(.... something using the field with internationalization )).join(', ')
})

SteRiccio (changes requested):

src/arena/taxonomy/index.js
ramirobg94: I don't see this repetition. Here I think that it is better to be explicit.

src/arena/taxonomy/index.js
```diff
@@ -0,0 +1,125 @@
...
+    label: '(Code), ScientificName, vernacularNames',
+  },
+  {
+    key: `${props.scientificName}.${props.vernacularNames}`,
+    value: [props.scientificName, props.vernacularNames],
+    label: 'ScientificName, vernacularNames',
+  },
+];
+
+export const exampleTaxonomy = {
```
SteRiccio: this one should be an exampleTaxon, not taxonomy...

ramirobg94: Fixed,

src/arena/taxonomy/index.js
SteRiccio: Do whatever you want then

SteRiccio: I was talking about the the entire taxonomyVisibleFieldsOptions object.
I mean that for every option the thing that you are considering is the fields and their order, for example [props.code, props.genus, props.family, props.scientificName]. When you build the options items, you can derive key and label properties from the value, so in the end it's better to define only an object with only the possible combinations, like:
{
codeScientificName: [props.code, props.scientificName],
scientificNameVernacularNames: [props.scientificName, props.vernacularNames],
...
}
and generate an "options" object derived by the one above... is it clear now?

ramirobg94: Please check the changes: https://github.com/openforis/arena-mobile/pull/415/commits/f231312eeb5e6848aede3934d938da51b73f7d44


In the future, I will be grateful if expressions such as "Is it clear now", "..." or "?!"; or categorical affirmations like "always", "never" or "a lot of" are avoided. These expressions could sound aggressive and inquisitive.

src/screens/Settings/Screens/Survey/Taxonomies/screen.js
```diff
@@ -61,47 +62,62 @@ const SettingsSurveyTaxonomies = () => {
...
-                  style={[
-                    styles.optionText,
-                    taxonomyVisibleFields === option.key &&
-                      styles.optionTextSelected,
-                  ]}>
-                  {option.label}
-                </Text>
-              </TouchableOpacity>
-            ))}
+            {Object.entries(taxonomyVisibleFieldsOptions)?.map(
```
SteRiccio: Object.entries always returns an array... no need for the optional chaining operator ?.

src/screens/Settings/Screens/Survey/Taxonomies/screen.js
```diff
@@ -61,47 +62,62 @@ const SettingsSurveyTaxonomies = () => {
...
+              (option, index) => {
+                console.log(option);
```
SteRiccio: please remove the console.log

src/screens/Settings/Screens/Survey/Taxonomies/screen.js
```diff
@@ -0,0 +1,143 @@
...
+            <Text style={baseStyles.textStyle.info}>
+              {t('Settings:survey.taxonomies.screen.description')}
+            </Text>
+            <Text style={baseStyles.textStyle.info}>
+              {t('Settings:survey.taxonomies.screen.info')}
+            </Text>
+
+            {Object.entries(taxonomyVisibleFieldsOptions)?.map(
+              (option, index) => {
+                console.log(option);
```
ramirobg94:
```suggestion
```

src/screens/Settings/Screens/Survey/Taxonomies/screen.js
```diff
@@ -0,0 +1,143 @@
...
+              {t('Settings:survey.taxonomies.screen.header')}
+            </Text>
+            <Text style={baseStyles.textStyle.info}>
+              {t('Settings:survey.taxonomies.screen.description')}
+            </Text>
+            <Text style={baseStyles.textStyle.info}>
+              {t('Settings:survey.taxonomies.screen.info')}
+            </Text>
+
+            {Object.entries(taxonomyVisibleFieldsOptions)?.map(
```
ramirobg94:
```suggestion
            {Object.entries(taxonomyVisibleFieldsOptions).map(
```

src/i18n/en/Settings.json
```diff
@@ -14,8 +14,25 @@
...
+      "screen": {
+        "header": "Taxonomies",
+        "description": "Select how do you wnat to see the taxonomies by default",
+        "info": "This value is going to be overwritten by the attribute configuration",
+        "fields": {
+          "code": "(Code)",
+          "genus": "Genus",
+          "family": "Family",
+          "scientificName": "ScientificName",
+          "vernacularNames": "VernacularNames"
```
ramirobg94:
```suggestion
          "scientificName": "Scientific Name",
          "vernacularNames": "Vernacular Names"
```

src/arena/taxonomy/index.js
SteRiccio: noted

SteRiccio (approved):

PR #440, opened 2023-06-07
persist-record

ramirobg94:


src/form/Attributes/Coordinate/Form/index.js
```diff
@@ -127,6 +127,9 @@ const Form = ({nodeDef}) => {
         defaultValue={String(newValue.x || node?.value?.x || '')}
         keyboardType={Platform.OS === 'ios' ? 'decimal-pad' : 'numeric'}
         textAlign="right"
+        editable={
+          nodeDef.props.allowOnlyDeviceCoordinate === true ? false : true
```
SteRiccio: you could have extracted this, something like:
const editable = !nodeDef.props.allowOnlyDeviceCoordinate

src/form/Attributes/common/BaseInput/Form/index.js
```diff
@@ -41,7 +41,10 @@ const Form = ({nodeDef, keyboardType = 'default'}) => {
   const handleClose = useCloseNode();
   const handleSubmit = useCallback(
     ({callback = handleClose} = {}) => {
-      const _newValue = newValue || String(node.value || '');
+      const _newValue =
+        newValue === '' || !Objects.isEmpty(newValue)
+          ? newValue
+          : String(node.value || '');
```
SteRiccio: this could be moved to some String util function, something like nullToDefault, or defaultIfNull, and you can call it in this way: 
StringUtils.defaultIfNull(newValue, node.value || '')

SteRiccio (approved):
just a couple of comments, feel free to follow the suggestions or not

PR #447, opened 2023-06-23
Theme setting

ramirobg94:


package.json
```diff
@@ -14,12 +14,11 @@
   },
   "dependencies": {
     "@miblanchard/react-native-slider": "^2.3.1",
-    "@openforis/arena-core": "^0.0.141",
-    "@react-native-async-storage/async-storage": "^1.18.1",
-    "@react-native-community/datetimepicker": "^6.7.5",
+    "@openforis/arena-core": "0.0.141",
```
persas, line 17: Why to remove the `^`? More relaxed deps avoid problems in the future

ramirobg94: In this case it is because the lastest arena-core package has a problem when updating the value of boolean attributes

persas (approved):

PR #451, opened 2023-06-28
fix multiple code, adjust styles to dark theme

ramirobg94:


src/form/Attributes/common/SearchableForm/List/index.js
```diff
@@ -20,16 +20,19 @@ const List = ({categoryItems, renderItem, searchText, nodes}) => {
...
+    const _categoryItemsUuidsFiltered = categoryItemsWithIndexToSearch
       .filter(
         categoryItem =>
-          (searchTextNormalized
+          searchTextNormalized
             ? categoryItem?.textForSearch.includes(searchTextNormalized)
-            : true) &&
-          !nodes.some(node => node?.value?.itemUuid === categoryItem.uuid),
+            : true,
+        /*DONT_SHOW_SELECTED !nodes.some(node => node?.value?.itemUuid === categoryItem.uuid) */
```
ramirobg94, line 30: This line is to remove the selected item when multiple, this is a behavior that maybe we want in the future but it is not required now

PR #454, opened 2023-07-17
Improve geolocation

ramirobg94:


src/form/Attributes/Coordinate/Form/GetLocation/index.js
```diff
@@ -59,7 +60,7 @@ const prepareItems = (location, {selectedSrs}) => {
     ].concat(
       Object.keys(location?.coords || {})
         .filter(key => !pareparedFields.includes(key))
-        .sort()
+        .sort((a, b) => (a > b ? 1 : a === b ? 0 : -1))
```
SteRiccio: keys are strings, if you use sort() without passing the custom sort function you should get the same result... isn´t it?

src/form/Attributes/Coordinate/Form/useGetLocation.js
```diff
@@ -98,27 +116,81 @@ const useGetLocation = () => {
...
+        forceLocationManager: true,
+        showLocationDialog: true,
+      },
+    );
+
+    return watcherId;
+  }, [hasLocationPermission, _setLocation]);
+
+  const getLocation = useCallback(() => {
+    const watcherId = _getLocation();
```
SteRiccio: _getLocation returns a Promise...

src/form/Attributes/Coordinate/Form/useGetLocation.js
```diff
@@ -98,27 +116,81 @@ const useGetLocation = () => {
...
+    return () => {
+      Geolocation.stopObserving();
+    };
+  }, []);
+
+  useEffect(() => {
+    return () => {
+      setLocation(false);
+    };
+  }, []);
```
SteRiccio, line 162: they can be combined into 1

src/state/form/selectors/index.js
```diff
@@ -383,6 +383,29 @@ const canAddNode = createCachedSelector(
...
 
+const generatePositionStringIndex = item => {
+  const {x, y} = item;
+  return `${(String(y) || '0').padStart(5, '0')}_${(String(x) || '0').padStart(
+    5,
+    '0',
+  )}`;
+};
+
+const getNodeDefAttributesSorted = ({nodeDefUuidsInEntity = [], isTable}) => {
```
SteRiccio: this should be named getNodeDefAttributeUuidsSorted actually, it doesn't return NodeDefs.

src/state/form/selectors/index.js
```diff
@@ -383,6 +383,29 @@ const canAddNode = createCachedSelector(
...
+  let formAttributesNodeDefs = isTable
+    ? nodeDefUuidsInEntity
+    : nodeDefUuidsInEntity
+        ?.sort((a, b) => {
+          const aIndex = generatePositionStringIndex(a);
+          const bIndex = generatePositionStringIndex(b);
+
+          return aIndex > bIndex ? 1 : -1;
+        })
+        .map(children => children?.i || children);
```
SteRiccio: when the renderType is form, it should be enough to sort the items first by y, then by x, with something like: sort((item1, item2) => item1.y - item2.y || item1.x - item2.x)

SteRiccio (approved):
more or less ok, just some concerns in useGetLocation, line 176, because _getLocation returns a Promise

SteRiccio:
please check the SonarCloud analysis too

src/form/Attributes/Coordinate/Form/GetLocation/index.js
ramirobg94: Sonar triggered a bug because the sort was empty

src/state/form/selectors/index.js
ramirobg94: no, because you could have columns

src/state/form/selectors/index.js
ramirobg94: true

src/form/Attributes/Coordinate/Form/useGetLocation.js
ramirobg94: Check the exmple on the docs
https://github.com/michalchudziak/react-native-geolocation/blob/master/example/src/examples/WatchPosition.tsx

src/form/Attributes/Coordinate/Form/GetLocation/index.js
SteRiccio: Really? The sort function is not mandatory. I think that someone can introduce more bugs by adding something that is done already by default...

src/form/Attributes/Coordinate/Form/useGetLocation.js
SteRiccio: I'm talking about your _getLocation function... Try to print out the value returned by that function

src/state/form/selectors/index.js
SteRiccio: Of course you can have columns, but I don't follow you.. can you provide an example of this please?

src/form/Attributes/Coordinate/Form/GetLocation/index.js
ramirobg94: Captura de pantalla 2023-07-17 a las 17 32 22

src/state/form/selectors/index.js
ramirobg94: 1 1
1 2
7 1
7 2
11 1

It is importat use the padStart to avoid the sort by string and not by number and also join both coordinates avoid to iterate 2 times

PR #463, opened 2023-08-03
Image quality

ramirobg94:


src/form/Attributes/File/components/Preview/component.js
```diff
@@ -27,12 +30,42 @@ export const FileImage = ({file}) => {
...
+    return StyleSheet.compose(
+      styles.imgContainer,
+      file?.meta?.width && file?.meta?.height
+        ? {
+            width:
+              (file.meta.width / Math.min(file.meta.width, file.meta.height)) *
+              MaxPreviewSize,
+            height:
+              (file.meta.height / Math.min(file.meta.width, file.meta.height)) *
+              MaxPreviewSize,
```
SteRiccio: `Math.min(file.meta.width, file.meta.height) / MaxPreviewSize` could be extracted to a value, something like resizeRatio or similar

SteRiccio (commented):
Is this or only for the image preview? images should be compressed to fit the max size specified in the file attribute definition too. In same cases compression is not needed (e.g. the user specifies 10MB as maximum size).

SteRiccio (approved):

PR #478, opened 2023-08-28
copy to clipboard

ramirobg94:


src/form/Attributes/Coordinate/Preview/index.js
```diff
@@ -26,6 +26,10 @@ const OpenInMap = ({nodeDef}) => {
     });
   };
 
+  if (!nodes[0]?.value?.x || !nodes[0]?.value?.y) {
```
SteRiccio, line 29: what if x or y is 0?

SteRiccio (approved):

src/form/Attributes/Coordinate/Preview/index.js
ramirobg94: true, but extremely weird! if you are there the waves are going to move you and the error is going to show some location, 
https://www.google.com/maps/place/0%C2%B000'00.0%22N+0%C2%B000'00.0%22E

PR #481, opened 2023-08-28
Lock record

ramirobg94:


src/screens/Form/components/EntityPage/components/common/EntityPanel/Navigation/component.js
```diff
@@ -148,58 +150,109 @@ const Next = ({parent}) => {
...
+      nodeDef: currentEntityNodeDef,
+      cycle,
+    });
 
-    return <NavigationButton nodeDef={nodeDef} />;
-  }
+    if (
+      !Objects.isEmpty(childrenIndex) &&
+      (Objects.isEmpty(parent) || parent === false || childrenIndex?.length > 0)
+    ) {
```
SteRiccio, line 163: you checked already that childrenIndex is not empty, so why do you verify if childrenIndex?.length > 0?

SteRiccio (approved):

src/screens/Form/components/EntityPage/components/common/EntityPanel/Navigation/component.js
ramirobg94: i will rmove them true

ramirobg94: I found one case in which makes a difference ->. I will check and simplify in other PR

PR #496, opened 2023-09-12
Adjustments validation

ramirobg94:


src/arena/record/index.js
```diff
@@ -15,10 +15,18 @@ const getKeyNodeAsString = ({
...
+  language,
+  showLabels = false,
 }) => {
   if (node?.value?.itemUuid) {
     const code = categoryItemIndex?.[node?.value?.itemUuid]?.props?.code;
-    return code || defaultString;
+    let label = false;
+    if (showLabels) {
+      label =
+        categoryItemIndex?.[node?.value?.itemUuid]?.props?.labels?.[language];
```
SteRiccio, line 26: this could be extracted into a constant, categoryItem, because it's used at line 22 too.
also, you are already checking at line 21 if node?.value?.itemUuid is defined, and it could be extracted inot another constant too, and arena-core NodeValues.getItemUuid could be used to get it from node

SteRiccio (approved):

src/arena/record/index.js
ramirobg94: true

PR #504, opened 2023-10-09
hide name

ramirobg94:


SteRiccio (approved):

src/form/Attributes/File/Preview/index.js
```diff
@@ -63,7 +63,7 @@ const FilePreview = ({node}) => {
       type="secondary"
       customContainerStyle={styles.bottonContainer}
       customTextStyle={styles.fileLabel}
-      label={node?.value?.fileName}
+      label={''}
```
SteRiccio, line 66: Maybe label could be optional in the Button component

PR #510, opened 2023-10-11
Error wrapper

ramirobg94:


src/form/Attributes/Coordinate/Form/index.js
```diff
@@ -138,18 +147,19 @@ const Form = ({nodeDef}) => {
       <GetLocation
         handleSaveLocation={handleSaveLocation}
         selectedSrs={selectedSrs}
+        surveySrsIndex={surveySrsIndex}
       />
       <CoordinateInput
         coordinateKey="x"
         handleUpdateValue={handleUpdateValue}
-        defaultValue={String(newValue.x || node?.value?.x || '')}
+        defaultValue={String(newValue?.x || node?.value?.x || '')}
```
SteRiccio, line 155: if x is 0 it will show an empty string there, isnt't it?

src/form/Attributes/Coordinate/Form/index.js
```diff
@@ -158,7 +168,7 @@ const Form = ({nodeDef}) => {
           coordinateKey="accuracy"
           handleUpdateValue={handleUpdateValue}
           defaultValue={String(
-            newValue.accuracy || node?.value?.accuracy || '',
+            newValue?.accuracy || node?.value?.accuracy || '',
```
SteRiccio, line 171: so as far as I understood, when newValue is null you pick the value from node: maybe it would be easier to have this check one time above and introduce some kind of valueUi constant, inialized with empty strings

SteRiccio (approved):
approved with a couple of comments

PR #511, opened 2023-10-25
Update versions

ramirobg94:


SteRiccio (approved):

src/form/Attributes/Coordinate/Form/GetLocation/index.js
```diff
@@ -102,7 +110,13 @@ const GetLocation = ({handleSaveLocation, selectedSrs, surveySrsIndex}) => {
           disabled={loading}
         />
       )}
-
+      {loading && !location && (
+        <ActivityIndicator
+          size="large"
+          color={styles?.colors?.primaryText}
+          style={styles.spinner}
+        />
```
SteRiccio, line 118: since you are using the same component with the same parameters inside ListRecords, maybe it could be extracted into a custome component that you can call SpinnerActivityIndicator or whatever you want

src/form/Attributes/Coordinate/Form/GetLocation/index.js
```diff
@@ -27,7 +27,15 @@ export const toFixedIfLongerAndNumber = (value, decimals = 3) => {
...
-const _sorter = (a, b) => (a > b ? 1 : a === b ? 0 : -1);
+const _sorter = (a, b) => {
+  if (a > b) {
+    return 1;
+  }
+  if (a === b) {
+    return 0;
+  }
+  return -1;
+};
```
SteRiccio, line 38: what's the difference with the default sort comparator?

src/form/Attributes/Taxonomy/Form/index.js
```diff
@@ -41,6 +41,19 @@ const getTextForSearch = (item, language) => {
   return normalizedString;
 };
 
+const generateKey = (item, textForSearch) => {
```
SteRiccio, line 44: what is this function for? please add some comments...

src/form/Attributes/Taxonomy/Form/index.js
```diff
@@ -41,6 +41,19 @@ const getTextForSearch = (item, language) => {
...
+  const vernacularNames = getVernacularNames({item});
+  const _textForSearch = textForSearch || item?.textForSearch || false;
+
+  const _key = _textForSearch
+    ? `${item.uuid}-${_textForSearch}-${Object.values(item?.vernacularNames)
+        .flatMap(i => i)
+        .map(i => i.uuid)}`
+    : `${item.uuid}-${vernacularNames.join(',')}`;
```
SteRiccio, line 52: maybe you can extract the part to get the vernacular name uuids and the final key can be:
`${item.uuid}-${vernacularNameUuids.join('-')}-${_textForSearch ?? ''}`

src/form/Attributes/Coordinate/Form/GetLocation/index.js
ramirobg94: sonar was complaining. First: It needs a sort function defined, then the ternary

src/form/Attributes/Taxonomy/Form/index.js
ramirobg94: yes maybe it is easier to read

src/form/Attributes/Coordinate/Form/GetLocation/index.js
ramirobg94: yes I will add other PR to unify this component

src/form/Attributes/Coordinate/Form/GetLocation/index.js
SteRiccio: I would have considered this just as a false positive...! you are compareing object keys (strings) with a function that is equal to the default sort comparator in the end...

PR #526, opened 2024-01-15
add uri to the file

ramirobg94:


SteRiccio:
Will the uri be included in the value sent to the server?

ramirobg94:
yes, should we remove that?

SteRiccio:
> yes, should we remove that?

it's not necessary in Arena... keep it in mobile if you need it but remove when you upload the records

PR #527, opened 2024-01-16
hide button

ramirobg94:


src/state/files/selectors.js
```diff
@@ -53,6 +55,7 @@ const getFilesByUuidNodeUuid = createCachedSelector(
 
 export default {
   getFiles,
+  getNumFiles,
```
SteRiccio, line 58: Where is it used?

SteRiccio (approved):
Approved with a comment

src/state/files/selectors.js
ramirobg94: future PR

PR #534, opened 2024-02-02
Multiple entity home

ramirobg94:
PR to create a multiple entity Home


src/screens/Form/components/BreadCrumbs/components/BreadCrumb/component.js
```diff
@@ -52,20 +52,58 @@ const BreadCrumb = ({nodeUuid}) => {
   const handleSelect = useCallback(() => {
     dispatch(
       formActions.setParentEntityNode({
-        node,
+        node: node || emptyNode,
       }),
     );
-  }, [node, dispatch]);
+  }, [node, emptyNode, , dispatch]);
```
SteRiccio: what is the double comma for?

src/screens/Form/components/BreadCrumbs/components/BreadCrumb/component.js
```diff
@@ -52,20 +52,58 @@ const BreadCrumb = ({nodeUuid}) => {
...
 
   return (
     <View style={styles.breadCrumbContainer}>
       <Pressable onPress={handleSelect} style={styles.container}>
-        <Label node={node} />
+        <Label node={node || emptyNode} />
```
SteRiccio, line 63: since "node || emptyNode" is repeating a couple of time, maybe it's better to extract it to a variable with a meaningful name

src/screens/Form/components/BreadCrumbs/components/CloseForm/component.js
```diff
@@ -13,25 +13,28 @@ import {useGetNumberOfErrors} from 'screens/Form/components/ValidationReport/hoo
 
 const EntitySelectorToggler = ({customStyle}) => {
   const dispatch = useDispatch();
-  const handleToggleEntitySelector = useCallback(() => {
-    dispatch(formActions.toggleEntitySelector());
-  }, [dispatch]);
+  const numberOfErrors = useGetNumberOfErrors();
+
+  const handleTogglePress = useCallback(
+    e => {
```
SteRiccio: SonarCloud should complain that the variable "e" is not used in this case...

SteRiccio (approved):
I'm quite curious to see how it works...

src/screens/Form/components/BreadCrumbs/components/BreadCrumb/component.js
ramirobg94: If they were 3 I would change that, but as it is "only" 2 cases I think that is not going to be worth it

PR #537, opened 2024-02-19
Unify home

ramirobg94:


src/arena-mobile-ui/components/SelectWithLabel/component.js
```diff
@@ -7,6 +7,7 @@ import {Label} from '../LabelsAndValues';
 import Select from '../Select';
 
 import _styles from './styles';
+import {pickerSelectStylesNeutral} from 'screens/Form/components/EntityPage/components/common/EntityPanel/EntityNodeSelector/styles';
```
SteRiccio: it's quite strange that this component under /components depends on a style defined in a component under /screens/... I would expect the opposite.

src/form/Attributes/Code/Preview/hooks/useCode.js
```diff
@@ -49,11 +49,11 @@ export const useCodeWithNode = (nodeDef, node) => {
     const expressionEvaluator = new RecordExpressionEvaluator();
     const items = node?.uuid ? nodeCategoryItems : nodeDefCategoryItems;
     if (!itemsFilter) {
```
SteRiccio, line 55: you could be more explicit here: 
`if (!items) return undefined`
and avoid the optional chaining operator in the following lines

src/screens/Home/components/Survey/components/SurveyDetail/component.js
```diff
@@ -1,72 +1,137 @@
...
+  const [showAll, setShowAll] = useState(false);
+  return (
+    <View style={styles.viewMoreContainer}>
+      <TextBase type="secondary" numberOfLines={showAll ? 0 : 2}>
+        {value}
+      </TextBase>
+      <Pressable
+        onPress={() => setShowAll(!showAll)}
+        style={styles.viewMoreButton}>
+        <TextBase type="bold">{showAll ? 'View less' : 'View more'}</TextBase>
```
SteRiccio: should these strings "view less/more" use localization?

src/screens/Home/components/Survey/components/SurveyDetail/component.js
```diff
@@ -1,72 +1,137 @@
...
+  if (labelSplit.length >= 2) {
+    return labelSplit[0][0] + labelSplit[1][0];
+  }
+
+  if (labelSplit.length >= 1) {
+    return labelSplit[0][0];
+  }
+
+  return labelSplit[0];
+};
```
SteRiccio: if I understand correctly, what you want to do here is to: 
1) split the text by space
2) take the first 3 parts
3) take the first letter of the parts
4) capitalize the letter
5) return a concatenation of the capitalized first letters  

so in practice it could be done with something similar to this:  

```
String(surveyLabel)
    .toUpperCase()
    .split(/\s+/)
    .slice(0, 3)
    .map(part => part[0])
    .join('')
```
that I think is more clear and if one day you decide to use 4 parts instead of 3, you can change the parameter passed to slice.

src/screens/Home/components/Survey/components/SurveyDetail/component.js
```diff
@@ -1,72 +1,137 @@
...
+      },
+    ];
+
+    if (numberOfCycles > 1) {
+      _info.push({
+        label: t('Home:survey.card.cycle'),
+        value: `${t('Home:survey.card.cycle_value', {
+          numberOfCycles,
+          cycle: `${String(Number(surveyCycle) + 1)} (${
+            surveyCycles[surveyCycle]?.dateStart
```
SteRiccio: is it possible to extact  surveyCycles[surveyCycle] into a variable, currentCycle or similar? it's quite difficult to understand what will be the resulting string generated in the lines 94-99

src/screens/Records/components/ListRecords/component.js
```diff
@@ -82,7 +83,7 @@ const ListRecords = ({selectedRecordUuid, setSelectedRecord}) => {
   );
 
   const loading = useSelector(recordSelectors.getIsGettingRemoteRecordsSummary);
-  if (recordUuids.length <= 0) {
+  if (recordUuids?.length <= 0) {
```
SteRiccio: when recordUuids is null or undefined, recordUuids?.length <= 0 will give false... isn't  it? and I think I've already commented in the past the array length <= 0... I give up!

SteRiccio (approved):
approved with some minor comments

src/arena-mobile-ui/components/SelectWithLabel/component.js
ramirobg94: Yep in fact this style should be on other 3rd place

src/screens/Records/components/ListRecords/component.js
ramirobg94: What do you like to say with "I give up"?

src/screens/Home/components/Survey/components/SurveyDetail/component.js
ramirobg94: right , I added the 3rd line to remove '-' and not needed spaces

String(surveyLabel)
    .toUpperCase()
    .replace(/[^a-zA-Z ]/g, '')
    .split(/\s+/)
    .slice(0, 3)
    .map(part => part[0])
    .join('')

src/screens/Home/components/Survey/components/SurveyDetail/component.js
ramirobg94: right

src/form/Attributes/Code/Preview/hooks/useCode.js
ramirobg94: I will check

src/screens/Home/components/Survey/components/SurveyDetail/component.js
ramirobg94: I will check

src/screens/Records/components/ListRecords/component.js
SteRiccio: just that I won't tell it to you anymore

SteRiccio (approved):

src/screens/Records/components/ListRecords/component.js
ramirobg94: Sorry, @SteRiccio. Could you share what "it" is? I am not getting this comment. Could you clarify?

ramirobg94: Ahhhh, now I understand!!!

It is the difference in style of the length <= 0 vs length === 0

that you commented a long time ago,
Sorry because, I didn't remember as I used to write with <= 0 just because the habit, 

I could change it to fit your preferences,

SteRiccio: I was joking above, I said that I won't tell it to you anymore, keep it as it is, it doesn't change anything in the end.

PR #539, opened 2024-03-06
Use mapsme to navigate

ramirobg94:


src/state/records/sagas/createRecord/index.js
```diff
@@ -24,6 +25,13 @@ const _createRecord = ({survey, user, cycle}) => {
     // _nodeRootUuid
     // _nodesByParentAndDef
     // _nodesByDef
+    appInfo: {
+      createdWith: {
+        appId: 'arena-mobile',
+        appVersion: DeviceInfo.getReadableVersion(),
+        appPlatform: DeviceInfo.getSystemName(),
+      },
+    },
```
ramirobg94: Please we have to check that it sends the expected info @SteRiccio

src/state/records/sagas/createRecord/index.js
```diff
@@ -6,9 +6,20 @@ import surveySelectors from 'state/survey/selectors';
...
 
 // extract this creators to arena-core
-const _createRecord = ({survey, user, cycle}) => {
+const _createRecord = async ({survey, user, cycle}) => {
+  let appVersion = '-';
+  let appPlatform = '-';
+
+  try {
+    appVersion = await DeviceInfo.getDeviceInfo().getReadableVersion();
+    appPlatform = await DeviceInfo.getDeviceInfo().getSystemName();
```
ramirobg94: https://www.npmjs.com/package/react-native-device-info


let systemName = DeviceInfo.getSystemName();
// iOS: "iOS" on newer iOS devices "iPhone OS" on older devices (including older iPad models), "iPadOS" for iPads using iPadOS 15.0 or higher.
// Android: "Android"
// Windows: ?

src/state/records/sagas/createRecord/index.js
```diff
@@ -24,6 +35,13 @@ const _createRecord = ({survey, user, cycle}) => {
     // _nodeRootUuid
     // _nodesByParentAndDef
     // _nodesByDef
+    appInfo: {
+      createdWith: {
+        appId: 'arena-mobile',
+        appVersion,
+        appPlatform,
```
SteRiccio: in the latest version of Arena I've renamed them to "version" and "platform", without the "app" prefix, is it ok for you or do you prefer to keep that prefix?

src/state/records/sagas/createRecord/index.js
SteRiccio: it doesn't matter too much now, it would be good to have some statistics from the server side about where the data comes from

src/infra/navigation/mapsme.js
```diff
@@ -0,0 +1,124 @@
+function transformName(s) {
+  return s
+    .split('')
+    .map(c => (c === ' ' ? '_' : c === '_' ? ' ' : c))
+    .join('');
+}
```
SteRiccio, line 6: not too much sure about what this does... a snake case? But I don't know why an empty space is converted to "_" while a "_" into an empty space, is it correct?

src/infra/navigation/mapsme.js
```diff
@@ -0,0 +1,124 @@
...
+    ) {
+      result += '%';
+      result += '0123456789ABCDEF'[c >> 4];
+      result += '0123456789ABCDEF'[c & 15];
+    } else {
+      result += s[i];
+    }
+  }
+  return result;
+}
```
SteRiccio, line 54: is it different from encodeURIComponent? and it's used only to encode a fixed name in your case, Location, am I wrong? so maybe it won't be necessary... also the constant arrays and strings could be moved out of the function into separate constants.

SteRiccio (approved):

src/infra/navigation/mapsme.js
ramirobg94: they do somehting wird I just copy and paste

PR #553, opened 2024-05-03
V1.44

ramirobg94:


src/i18n/de/Settings.json
```diff
@@ -109,6 +109,9 @@
     "hasToLockRecordsWhenLeave": {
       "title": "Datensätze sperren, wenn Formular verlassen wird"
     },
+    "hasToShowNotRelevantOnNavigationTree": {
```
SteRiccio, line 112: We agreed that the option was not needed, am I wrong? Not relevant nodes, hidden when not relevant, and entities marked as hidden in mobile should be hidden, without the need of the option

SteRiccio (changes requested):

SteRiccio (approved):

PR #557, opened 2024-05-10
Fixes

ramirobg94:


src/screens/Surveys/components/common/SurveyCard/component.js
```diff
@@ -37,26 +38,36 @@ const SurveyCard = ({
...
       <View style={styles.infoContainer}>
         <View style={styles.payload}>
-          <View style={styles.labelCotainer}>
-            <TextBase type="bold" size="l">
-              {survey.props?.labels?.[survey?.props?.languages?.[0]]}
-            </TextBase>
+          <View style={styles.visualDescriptionContainer}>
+            <VisualDescription
+              surveyLabel={
+                survey?.props?.labels?.[survey?.props?.languages?.[0]]
```
SteRiccio: this could be extracted into a constant, it's used at line 52 too and if one day you want to show the label in the device language (if available) it would be easier to change

SteRiccio (approved):

PR #570, opened 2024-07-05
improvements

ramirobg94:


src/analytics/events/uploadData/handlePrepareRecordsData/index.js
```diff
@@ -0,0 +1,16 @@
...
+  continue: defaultEvent({
+    type: 'UploadData handlePrepareRecordsData Continue',
+  }),
+  success: defaultEvent({
+    type: 'UploadData handlePrepareRecordsData Success',
+  }),
+  error: defaultEvent({
+    type: 'UploadData handlePrepareRecordsData Error',
+  }),
+};
```
SteRiccio, line 16: these events look always the same, only the suffix of the type changes, in this case "UploadData handlePrepareRecordsData". Can it be more generic?

src/state/records/sagas/getRemoteRecordsSummary/index.js
```diff
@@ -39,18 +40,23 @@ function* handleGetRemoteRecordsSummary() {
...
 
-    const recordsSummaryByRecordUuid = list.reduce((acc, recordSummary) => {
-      acc[recordSummary.uuid] = recordSummary;
-      return acc;
-    }, {});
+    let recordsSummaryByRecordUuid = {};
+    if (!Objects.isEmpty(recordsSummary?.list)) {
+      const {list} = recordsSummary;
+
+      recordsSummaryByRecordUuid = (list || []).reduce((acc, recordSummary) => {
```
SteRiccio, line 48: list is not empty at this point, you were already checking if list is not empty before, at line 45

SteRiccio (approved):

PR #572, opened 2024-07-19
Upload one by one

ramirobg94:


src/form/common/Form/component.js
```diff
@@ -63,7 +66,7 @@ export const RenderFormContainer = () => {
     return (
       <KeyboardAvoidingView
         style={styles.container}
-        behavior={Platform.OS === 'ios' ? 'padding' : 'height'}
+        behavior={keyboardAvoidingBehavior}
         enabled>
```
SteRiccio, line 70: maybe it was simpler to wrap KeyboardAvoidingView into your own component, so that you don't have to remember to pass a custom "behaviour" every time you use it

SteRiccio (approved):

PR #606, opened 2024-11-26
Fix multiple entity home keys visibility 

### LINKED ISSUE #599: multiple entity summary view: strange behaviour after item deletion
it happens when the entity has more than one key attributes; first only one key appears in the table; when an item is deleted, both keys appear.

https://github.com/user-attachments/assets/96e6fd4a-993c-4324-b59f-bffe355fd234

ramirobg94:
resolves #599


src/state/form/selectors/index.js
```diff
@@ -423,6 +423,10 @@ const getFormAttributesNodeDefs = createSelector(
     formAttributesNodeDefs = formAttributesNodeDefs
       ?.map(nodeDefUuid => nodeDefsByUuid[nodeDefUuid])
       .filter(_nodeDef => {
+        if (NodeDefs.isKey(_nodeDef)) {
+          return true;
+        }
+
```
SteRiccio: the check Objects.isEmpty(_nodeDef) could be done first, then all other checks.
Please also notice that there are helper methods in Nodes to extract other properties, like Nodes.isHidden, Nodes.isHiddenInMobile.

SteRiccio (changes requested):

src/state/form/selectors/index.js
ramirobg94: if (Objects.isEmpty(_nodeDef)) { return false; }
if (NodeDefs.isKey(_nodeDef)) { return true; }
if (_nodeDef?.props?.hidden) { return false; }
if (NodeDefs.isHiddenInMobile(cycle)(_nodeDef)){ return false; }

src/state/form/selectors/index.js
```diff
@@ -423,6 +423,10 @@ const getFormAttributesNodeDefs = createSelector(
...
         if (Objects.isEmpty(_nodeDef)) {
           return false;
         }
```
ramirobg94:
```suggestion
        if (Objects.isEmpty(_nodeDef)) {
          return false;
        }
```
```suggestion
```

src/state/form/selectors/index.js
```diff
@@ -423,6 +423,10 @@ const getFormAttributesNodeDefs = createSelector(
     formAttributesNodeDefs = formAttributesNodeDefs
       ?.map(nodeDefUuid => nodeDefsByUuid[nodeDefUuid])
       .filter(_nodeDef => {
+        if (NodeDefs.isKey(_nodeDef)) {
+          return true;
+        }
```
ramirobg94:
```suggestion
        if (Objects.isEmpty(_nodeDef)) {
          return false;
        }
        
        if (NodeDefs.isKey(_nodeDef)) {
          return true;
        }
```

SteRiccio (approved):

src/state/form/selectors/index.js
```diff
@@ -426,13 +426,16 @@ const getFormAttributesNodeDefs = createSelector(
...
+
         if (_nodeDef?.props?.hidden) {
           return false;
         }
 
-        const layoutProps = NodeDefs.getLayoutProps(cycle)(_nodeDef);
-
-        if (layoutProps?.hiddenInMobile === true) {
+        if (NodeDefs.isHiddenInMobile(cycle)(_nodeDef)) {
```
SteRiccio, line 438: NodeDefs.isHidden can be used here instead of _nodeDef?.props?.hidden

src/state/form/selectors/index.js
```diff
@@ -426,13 +426,16 @@ const getFormAttributesNodeDefs = createSelector(
         if (Objects.isEmpty(_nodeDef)) {
           return false;
         }
+
+        if (NodeDefs.isKey(_nodeDef)) {
+          return true;
+        }
+
         if (_nodeDef?.props?.hidden) {
```
ramirobg94:
```suggestion
        if (NodeDefs.isHidden(_nodeDef)) {
```

SteRiccio (approved):

PR #607, opened 2024-11-27
Files with computed name

ramirobg94:


src/form/Attributes/File/Cell/index.js
```diff
@@ -5,7 +5,14 @@ import BaseCell from 'form/Attributes/common/Base/Cell';
 
 const BaseValuesRenderer = ({nodes}) => (
   <TextBase numberOfLines={1}>
-    {nodes?.map(node => node?.value?.fileName).join(', ')}
+    {nodes
+      ?.map(node => {
+        const {fileName: fileNameUuid, fileNameCalculated} =
+          node?.value?.fileName;
```
SteRiccio, line 11: this one looks wrong, it should be:
const {fileName: fileNameUuid, fileNameCalculated} =
          node?.value ?? {}

src/form/Attributes/File/Preview/index.js
```diff
@@ -86,7 +86,7 @@ const NodeValueRender = ({node = false, nodeDef}) => {
   return (
     <>
       <Container>
-        {node?.value ? (
+        {node?.value && node?.value?.fileName ? (
```
SteRiccio, line 89: it could be simply node?.value?.fileName

src/form/Attributes/File/hooks/useFile.js
```diff
@@ -59,7 +65,7 @@ const useFile = ({nodeDef, node, isImage = false}) => {
   const deleteFile = useCallback(() => {
     handleDelete({
       node,
-      label: node?.value?.fileName,
+      label: node?.value?.fileNameCalculated,
```
SteRiccio, line 68: in this case, if fileNameCalculated is missing it should default to something else

SteRiccio (changes requested):

SteRiccio:
also please check the SonarCloud issues

src/form/Attributes/File/Preview/index.js
SteRiccio: there are also helper methods in arena-core for this, e.g. NodeValues.getFileName

